PR #7, opened 2017-02-15
Test Vibe.d compatibility with Travis

wilzbach:
The idea is to run all Vibe.d tests with every change on vibe-core to ensure that no "random breakage" happens.

### DIFF
--- a/travis-ci.sh
+++ b/travis-ci.sh
@@ -2,6 +2,21 @@
 
 set -e -x -o pipefail
 
+clone() {
+    local url="$1"
+    local path="$2"
+    for i in {0..4}; do
+        if git clone "$url" "$path" "${@:3}"; then
+            break
+        elif [ $i -lt 4 ]; then
+            sleep $((1 << $i))
+        else
+            echo "Failed to clone: ${url}"
+            exit 1
+        fi
+    done
+}
+
 DUB_FLAGS=${DUB_FLAGS:-}
 
 # test for successful release build
@@ -33,3 +48,8 @@ if [ ${RUN_TEST=1} -eq 1 ]; then
         fi
     done
 fi
+
+# test compatibility with current Vibe.d
+clone https://github.com/rejectedsoftware/vibe.d vibe.d --depth 1
+sed 's/dependency "vibe-core" version="~(.*)"/dependency "vibe-core" path=".."/' -i vibe.d/core/dub.sdl
+(cd vibe.d && VIBED_DRIVER="vibe-core" BUILD_EXAMPLE=1 RUN_TEST=1 ./travis-ci.sh)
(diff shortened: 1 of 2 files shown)


travis-ci.sh
```diff
@@ -7,7 +7,7 @@ dub build -b release --compiler=$DC -c $CONFIG
 
 # test for successful 32-bit build
 if [ "$DC" == "dmd" ]; then
-	dub build --arch=x86 -c $CONFIG
+    dub build --arch=x86 -c $CONFIG
```
wilzbach: The file was mixed with tabs & whitespace -> I unified to spaces.

s-ludwig:
Good idea! Instead of `add-local`+`~master`, I'd just use `path=".."` in the dependency to avoid the warnings from DUB and to make sure that no surprises happen on non-master branches (actually not sure how it will behave).

travis-ci.sh
```diff
@@ -25,3 +25,9 @@ if [ ${RUN_TEST=1} -eq 1 ]; then
         dub --temp-build --compiler=$DC --single $ex # --override-config vibe-core/$CONFIG
     done
 fi
+
+# test compatibility with current Vibe.d
+dub add-local .
+git clone https://github.com/rejectedsoftware/vibe.d
```
s-ludwig: `--depth 1` makes sense here to reduce the network traffic a bit (mainly skips the old Windows library binaries for OpenSSL and libevent).

wilzbach:
Hmm with DMD 2.073 it errors as follows:

> Error: Error writing file '.dub/build/vibe-d-test-vibe-core-unittest-linux.posix-x86_64-dmd_2073-D0D1168615CCAD2773D24ECEDF3320F7/vibe-d-test-vibe-core.o'

It's a bit strange because I can't reproduce this locally and on the [previous run](https://travis-ci.org/vibe-d/vibe-core/builds/202116597) only the "epoll" config build ran into this error.
Could it be that DMD runs out of memory (e.g. due to the `ctRegex` regression?)

s-ludwig:
I get that error every other build for vibe.d itself since a while. Not sure what the real issue is - I suspected an out of disk space error, but clearing intermediate build results between the different builds didn't really help either. Maybe this should be left to the minimum of `dub build --combined` for now. I'll add all vibe:core tests to this repo anyway, so that should already cover a lot of ground.

wilzbach:
> I get that error every other build for vibe.d itself since a while. Not sure what the real issue is - I suspected an out of disk space error, but clearing intermediate build results between the different builds didn't really help either.

Let's have a look whether it's more stable now :)

s-ludwig:
Closing since Travis isn't available anymore.

PR #10, opened 2017-03-22
Allow removal of the default logger

wilzbach:
Last week I discovered an imho important use case of `vibe.core.log`: I wanted to display more information on the default console logger (file name, line number, ...) and thought that I can just implement my own logger and deregister the default one. However, unfortunately deregistering the default logger isn't possible as no reference to it is exposed to the user.

An alternative to exposing `getLoggers`, would be to have a `deregisterAll` method.

Btw what is the reason for commenting `string mod = __MODULE__, string func = __FUNCTION__,`?
Did it add too much bloat to the compiled binary?

### DIFF
--- a/source/vibe/core/log.d
+++ b/source/vibe/core/log.d
@@ -754,7 +754,10 @@ private {
 	shared(FileLogger) ss_stdoutLogger;
 }
 
-private shared(Logger)[] getLoggers() nothrow @trusted { return ss_loggers; }
+/**
+Returns a list of all registered loggers.
+*/
+shared(Logger)[] getLoggers() nothrow @trusted { return ss_loggers; }
 
 package void initializeLogModule()
 {


s-ludwig:
It's unfortunately not really inferable from its name due to historic reasons, but `setLogLevel(LogLevel.none);` will actually disable (just) the default logger. Removing it altogether is of course slightly more efficient.

I'm pretty sure that the module/function parameters were commented out due to template bloat, but it would make sense to actually measure things with a minimal wrapper (moving most of `log()` into a separate function that takes the arguments as runtime parameters).

s-ludwig:
Also would be interesting to make the file logger configurable with a custom format string.

s-ludwig:
Actually, just being able to query the logger sounds like a logical thing to enable. It kind of makes the thread-safety issue in the log module worse (only allowed to call register/unregister logger before any thread was started), but that can be improved as a whole as part of a different change. I'll merge this as-is for now.

wilzbach:
>  It kind of makes the thread-safety issue in the log module worse (only allowed to call register/unregister logger before any thread was started), but that can be improved as a whole as part of a different change. I'll merge this as-is for now.

Totally forgot about this PR. Thanks!

PR #14, opened 2017-05-29
Improve Travis CI setup

wilzbach:
Analog to https://github.com/rejectedsoftware/vibe.d/pull/1772, https://github.com/dlang/dub/pull/1138, https://github.com/dlang/dub-registry/pull/214, https://github.com/etcimon/botan/pull/30 and https://github.com/etcimon/libasync/pull/77

This is also intended to be used with daily crons. We could have a dedicated slack channel or mailing list for such errors resulting from `dmd-nightly`?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,13 +5,17 @@ d:
   # this way the overall test time gets cut down (GDC/LDC are a lot
   # slower tham DMD, so they should be started early), while still
   # catching most DMD version related build failures early
-  - dmd-2.073.0
+  - dmd-2.073.2
   - dmd-2.070.2
-  - ldc-1.1.0
+  - ldc-1.2.0
   - ldc-1.0.0
   - dmd-2.072.2
   - dmd-2.071.2
+  - dmd
   - dmd-beta
+  - dmd-nigthly
+  - ldc-beta
+  - ldc
 
 env:
     - CONFIG=select
@@ -21,8 +25,15 @@ env:
 matrix:
   allow_failures:
     - env: CONFIG=libasync
-    - d: dmd-beta
 
 script: ./travis-ci.sh
 
 sudo: false
+
+notifications:
+  slack:
+    rooms:
+      - dlang:3mRWPCAIEVg3barvUoCzgnN6
+    on_pull_requests: false
+    on_success: change
+    on_failure: always


.travis.yml
```diff
@@ -5,24 +5,23 @@ d:
...
+  - ldc
 
 env:
     - CONFIG=select
     - CONFIG=epoll
     - CONFIG=libasync
 
-matrix:
-  allow_failures:
-    - env: CONFIG=libasync
```
s-ludwig: This needs to stay for the time being (libasync driver not implemented yet).

wilzbach:
> A filtered channel would definitely be a good idea (would have to filter failures that fail only on the nightly/beta), because at least for the repositories that I watch, failures will most likely get lost in the general Travis noise.

Okay I setup a simple slack integration that should post to the "vibe-d-ci" channel when `master` build fail.

https://docs.travis-ci.com/user/notifications/#Configuring-slack-notifications
https://dlang.slack.com/messages/C5KC27YCB/convo/C0JP1U9L5-1495545485.530804

I'm not sure how easy it is to filter only for nightly/beta, but Martin & I wanted to increase the Travis/Dub stability anyways, so it might be good to have an eye on what random failures occur.
However, as it's an experiment I only enabled it for this repo, which is currently low traffic and thus will in the worst case at most trigger one notification a day.

s-ludwig:
Since it doesn't look like it would be possible to get slack notifications in case the nightly tests are marked as `allowedFailures` and the nightly still fails, I'm not sure if this is a viable approach. Having the package  appear in a broken state is definitely not an option and I'd expect this to happen from time to time.

wilzbach:
> Since it doesn't look like it would be possible to get slack notifications in case the nightly tests are marked as allowedFailures and the nightly still fails, I'm not sure if this is a viable approach. Having the package appear in a broken state is definitely not an option and I'd expect this to happen from time to time.

Yup I agree, we might want to setup a meta package that basically git clones Vibe.d and builds it, s.t. the failures from nightly don't make the package appear in a broken state.
OTOH I really hope that we can rectify Martin's project tester which is on a [good way currently](https://github.com/dlang/ci/pull/54), s.t. maybe we don't need this at all.

PR #18, opened 2017-06-28
Use a thread local allocator instead of processAllocator in HashMap.

s-ludwig:
This is required since processAllocator returns an ISharedAlloactor since 2.075.0. Because HashMap is operating only thread-local, this should generally be safe.


s-ludwig:
Fortunately the remaining changes are more or less search and replace. But it's good that it came up on the NG, as this had already disappeared from by radar.

PR #19, opened 2017-06-30
Add initial CodeCov setup

wilzbach:
Similarly to https://github.com/rejectedsoftware/vibe.d/pull/1804

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,0 +1,14 @@
+# Documentation: https://github.com/codecov/support/wiki/codecov.yml
+
+coverage:
+  precision: 3
+  round: down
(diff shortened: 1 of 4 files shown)


wilzbach:
It seems like the initial effort has worked: https://codecov.io/gh/vibe-d/vibe-core/tree/786c1fce0399effd18d29a1199062bb663eaf980/source/vibe

Of course, in a next step the coverage could be enabled for the examples as well.

s-ludwig:
> for the examples as well

Or at least for the high-level tests. BTW, are the .lst files actually updated incrementally, or do they get overwritten for each compiler invocation?

PR #21, opened 2017-07-02
Remove all trailing whitespace + add check

wilzbach:
Analogous to https://github.com/rejectedsoftware/vibe.d/pull/1800

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,8 @@ d:
 env:
     - CONFIG=select
     - CONFIG=epoll
-    - CONFIG=libasync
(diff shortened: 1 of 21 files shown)


wilzbach:
Thanks for picking this up. FYI as mentioned in the git commit message, there is a simple way to catch all occurrences: `sed 's/[ \t]*$//' -i **/*.d`

wilzbach:
There seems to be a problem with AppVeyor:

image

s-ludwig:
Appears to be going unusually slow, but from time to time it starts a new build. Anyway, the test doesn't event run on Windows, so I'll just override-merge.

PR #22, opened 2017-07-02
Remove deprecated stdc import

wilzbach:
Analogous to https://github.com/rejectedsoftware/vibe.d/pull/1811

> This was probably not seen because no developers actively compiles Vibe.d on Windows and looked at the deprecation warnings.

See also: https://github.com/dlang/phobos/pull/5532

### DIFF
--- a/source/vibe/core/core.d
+++ b/source/vibe/core/core.d
@@ -1264,7 +1264,7 @@ shared static this()
 		static if (need_wsa) {
 			logTrace("init winsock");
 			// initialize WinSock2
-			import std.c.windows.winsock;
+			import core.sys.windows.winsock2;
 			WSADATA data;
 			WSAStartup(0x0202, &data);
 
--- a/source/vibe/core/net.d
+++ b/source/vibe/core/net.d
@@ -34,7 +34,7 @@ NetworkAddress resolveHost(string host, AddressFamily address_family = AddressFa
 NetworkAddress resolveHost(string host, ushort address_family, bool use_dns = true)
 {
 	import std.socket : parseAddress;
-	version (Windows) import std.c.windows.winsock : sockaddr_in, sockaddr_in6;
+	version (Windows) import core.sys.windows.winsock2 : sockaddr_in, sockaddr_in6;
 	else import core.sys.posix.netinet.in_ : sockaddr_in, sockaddr_in6;
 
 	enforce(host.length > 0, "Host name must not be empty.");


wilzbach:
image

FYI: this means that the test suite isn't deterministic (it yields different coverage output).
At least here, the line in question is:

image

Basically there are basically two solutions:
- Add a test that always covers this line (usually quite tricky for concurrency)
- Ignore this line (see https://github.com/dlang/phobos/pull/5375 for a very simple trick)

s-ludwig:
Hm, is anything still missing for the auto-merge feature or dlang-bot in general?

>  test suite isn't deterministic

Interesting. That particular line looks like it should always be called at least once (on the first call to `acquireThreadWaiter`), though. Just to be sure to be on the right track, do we know if DMD's coverage feature is properly thread-safe?

wilzbach:
>  Just to be sure to be on the right track, do we know if DMD's coverage feature is properly thread-safe?

Looks `__gshared` to me:

https://github.com/dlang/druntime/blob/master/src/rt/cover.d

Will have a dive into DMD as well.

wilzbach:
>  s-ludwig added the auto-merge label 19 minutes ago

Hmm I am still not receiving no hook events.
Is the hook configured correctly here?

See also: https://github.com/dlang-bots/dlang-bot#using-the-dlang-bot-for-your-project

wilzbach:
> Will have a dive into DMD as well.

Looks like it's not synchronized:

```d
elem *e;
e = el_ptr(m.cov);
e = el_bin(OPadd, TYnptr, e, el_long(TYuint, linnum * 4));
e = el_una(OPind, TYuint, e);
e = el_bin(OPaddass, TYuint, e, el_long(TYuint, 1));
```

https://github.com/dlang/dmd/blob/master/src/ddmd/toir.d#L62

wilzbach:
> @s-ludwig s-ludwig added auto-merge and removed auto-merge labels 5 minutes ago

Now I see something in the logs :tada: 

```
» 5 Jul 2017 13:24:40.563  176 <190>1 2017-07-05T11:24:40.221700+00:00 app web.1 - - [FD264F5E:FD276D5E dbg] [github/pull_request](vibe-d/vibe-core/22): action=labeled
5 Jul 2017 13:24:40.902  206 <190>1 2017-07-05T11:24:40.534992+00:00 app web.1 - - [FD264F5E:FD26F15E WRN] PUT https://api.github.com/repos/vibe-d/vibe-core/pulls/22/merge failed;  Not Found 404.
» 5 Jul 2017 13:24:40.982  236 <190>1 2017-07-05T11:24:40.535007+00:00 app web.1 - - [FD264F5E:FD26F15E WRN] {"message":"Not Found","documentation_url":"https://developer.github.com/v3/pulls/#merge-a-pull-request-merge-button"}
```

Not sure yet, why GH doesn't like this.

wilzbach:
> Looks like it's not synchronized:

On a second thought, this shouldn't matter, because even if both threads try to increase the coverage of line X, we would still end up with 1 (i.e. not 0) in the worst case.

s-ludwig:
Okay, so then I'd suspect that to be the responsible. Adding synchronization is one possibility, or, probably the better approach, making the coverage data thread-local and adding some synchronized merge code in the thread cleanup code. Guess that should be a ticket against Druntime then.

I forgot to add a webhook for the vibe-d organization indeed. It's now there and lists some successful post requests, no reaction after re-applying the auto-merge label, though. The vibe.d repository lists failed post requests: `object.Exception@source/dlangbot/github_api.d(284): Hook signature mismatch` (does that mean I mistyped the secret key?)

wilzbach:
> object.Exception@source/dlangbot/github_api.d(284): Hook signature mismatch (does that mean I mistyped the secret key?)

Yeah, that's strange, especially as mentioned above, the bot now does receive events.
Maybe it's the easiest if - until we have have fixed this - you add me to the project, s.t. I can try re-applying `auto-merge` myself?

s-ludwig:
> On a second thought, this shouldn't matter, because even if both threads try to increase the coverage of line X, we would still end up with 1 (i.e. not 0) in the worst case.

You are probably right, unless the compiler emits strange code that doesn't operate on 32-bit boundaries, which doesn't really make sense here. So then this remains mysterious. I can take a deeper look when I get some more time.

s-ludwig:
> Maybe it's the easiest if - until we have have fixed this - you add me to the project, s.t. I can try re-applying auto-merge myself?

Sent an invitation.

wilzbach:
> Sent an invitation.

Thanks. I think I have identified the problem. The @dlang-bot isn't a member yet.
@MartinNowak would you be so kind to accept the invitation (if there's one)?

### Without authorization

```
> curl -H "Content-Type: application/json" -X PUT -d '{"commit_message":"Remove deprecated stdc import\nmerged-on-behalf-of: Sebastian Wilzbach <wilzbach@users.noreply.github.com>", "merge_method":"merge","sha":"f8af374f8613bdd555ecda4a4b7477d6aff3cb30"' https://api.github.com/repos/vibe-d/vibe-core/pulls/22/merge        
```

```json
{
  "message": "Not Found",
  "documentation_url": "https://developer.github.com/v3/pulls/#merge-a-pull-request-merge-button"
}
```

### With authorization

```
>curl -H "Content-Type: application/json" -H "Authorization: token XXX" -X PUT -d '{"commit_message":"Remove deprecated stdc import\nmerged-on-behalf-of: Sebastian Wilzbach <wilzbach@users.noreply.github.com>","merge_method":"merge","sha":"04b3575c14dc7ad9971e19f153f3e3d712c1bdde"' https://api.github.com/repos/vibe-d/vibe-core/pulls/22/merge
```

```json
{
  "sha": "cb169b850231891df90111ff1469a96e385080f4",
  "merged": true,
  "message": "Pull Request successfully merged"
}
```

### Added this to the testsuite

I tried to reproduce this locally and added the payloads to the Dlang-Bot's test suite:

-> https://github.com/dlang-bots/dlang-bot/pull/126

More test data doesn't hurt :)

s-ludwig:
> Thanks. I think I have identified the problem. The @dlang-bot isn't a member yet.
@MartinNowak would you be so kind to accept the invitation (if there's one)?

Ah, wait a minute! dlang-bot is in the team, but the team was not authorized for this repository (going to do the same for eventcore now, too).

PR #25, opened 2017-07-18
Implement typed cross-task channels

s-ludwig:
Sketches a Channel class, similar to Go's channels, to enable CSP style programming. They are also more efficient than the dynamically typed message queues used by `std.concurrency`.

The basic functionality is there, but various design questions are still unresolved.


s-ludwig:
The core semantics are clear now and the implementation should be ready for some real world use. Performance optimzation and additional flexibility (waiting on multiple channels, unbuffered channels) can be added on top later on.

PR #27, opened 2017-07-21
Convert Waitable from struct to pure template to avoid heap closures.

s-ludwig:
A first profiling run showed that every call to `asyncAwait` produced three to four heap closures, due to the way the different callbacks were wrapped. This change eliminates all of these to make room for more performance tuning.

### DIFF
--- a/source/vibe/core/core.d
+++ b/source/vibe/core/core.d
@@ -1010,14 +1010,17 @@ struct FileDescriptorEvent {
 
 		assert((which & m_trigger) == Trigger.read, "Waiting for write event not yet supported.");
 
-		Waitable!(IOCallback,
+		bool got_data;
+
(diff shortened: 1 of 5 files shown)


s-ludwig:
Going to drop 2.070.0 support to sidestep those linker errors.

s-ludwig:
Hitting https://github.com/DigitalMars/optlink/issues/18 here. Since issues like these often happen in a seemingly random fashion, I'm inclined to add the Optlink test runs to `allowed_failures`. (@WalterBright, FYI, if you have the time to look at it, this would provide a relatively self-contained reproduction case)

s-ludwig:
This also leads to 20% performance improvement for the bench-dummy-http-server example alone.

PR #31, opened 2017-08-28
fix deadlocks, caused by second m_locks increment

Boris-Barboris:
Got some nasty deadlocks in connection pool because of this.


Boris-Barboris:
@s-ludwig sorry, do you mind having a look? This one might actally hurt people if left untouched.
There is a second, unneeded semaphore counter  increment right after the one inside tryLock() call inside if condition. I found this during siege tests of Psql connection pool, after this change everything works as expected.

s-ludwig:
Thanks! I didn't notice this, because the default `ConnectionPool` limit is `uint.max`. There really need to be some basic unit tests for `TaskSemaphore` in the future.

PR #38, opened 2017-10-24
Update to more universal inout vs. const

schveiguy:
Eliminates an unnecessary overload.


schveiguy:
Closing, see related PR for explanation.
https://github.com/vibe-d/vibe.d/pull/1962

PR #41, opened 2017-11-11
tcp connect timeout

Boris-Barboris:
This is a sibling of https://github.com/vibe-d/eventcore/pull/28

### DIFF
--- a/source/vibe/core/net.d
+++ b/source/vibe/core/net.d
@@ -195,18 +195,26 @@ TCPConnection connectTCP(NetworkAddress addr, NetworkAddress bind_address = anyA
 		scope uaddr = new RefAddress(addr.sockAddr, addr.sockAddrLen);
 		scope baddr = new RefAddress(bind_address.sockAddr, bind_address.sockAddrLen);
 
-		auto result = asyncAwait!(ConnectCallback,
+		bool cancelled;
+		StreamSocketFD sock;
+		ConnectStatus status;
+
+		alias waiter = Waitable!(ConnectCallback,
 			cb => eventDriver.sockets.connectStream(uaddr, baddr, cb),
 			(ConnectCallback cb, StreamSocketFD sock_fd) {
+				cancelled = true;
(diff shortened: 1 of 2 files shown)


source/vibe/core/net.d
```diff
@@ -193,11 +195,15 @@ TCPConnection connectTCP(NetworkAddress addr, NetworkAddress bind_address = anyA
...
-			//cb => eventDriver.sockets.cancelConnect(cb)
-		);
+			cb => eventDriver.sockets.connectStream(uaddr, baddr, cb),
+			(ConnectCallback cb, StreamSocketFD sock_fd) {
+				eventDriver.sockets.cancelConnectStream(sock_fd);
+			}
+		)(timeout);
+
+		if (result[1] == ConnectStatus.cancelled)
+			result[1] = ConnectStatus.timeout;
```
s-ludwig: `asyncAwaitUninterruptible` could be changed to `asyncAwait`, now that cancellation is possible. The `if` condition above would then have to be `(!result.completed)`, wich will also work with the updated callback semantics for the eventcore PR.

Boris-Barboris:
apparently, linker hanged (

s-ludwig:
It's apparently hitting the same issue as the one worked around in 72540d7eaf1f3e1e0b92925d0660b78744835414. I can test this locally and try to find a workaround later.

Boris-Barboris:
*unobtrusive fart*

Boris-Barboris:
Ah, ok, so 1cadc3077bc5c1b05ac10b08c5cdfab236b53c4d is needed, got it.    
There also OOM crash in https://ci.appveyor.com/project/s-ludwig/vibe-core/build/1.0.205/job/njsty59hfqcddue9

s-ludwig:
#41 should pass now, it's the branch rebased on top of master and includes the `asyncAwaitAny` fix, which is still necessary for Optlink and DMD 2.076.x and below. Also got no OOM error so far, that could just have been a thing on that particular instance.

Boris-Barboris:
cool, thank you

PR #43, opened 2017-11-21
Use the stable stdx-allocator DUB package

wilzbach:
As discussed before, `stdx-allocator` is a frozen version of std.experimental.allocator to avoid the frequent breakages. 

See also: https://github.com/dlang-community/discussions/issues/27

### DIFF
--- a/dub.sdl
+++ b/dub.sdl
@@ -5,6 +5,7 @@ copyright "Copyright © 2016, rejectedsoftware e.K."
 license "MIT"
 
 dependency "eventcore" version="~>0.8.18"
+dependency "stdx-allocator" version="~>2.77.0"
 
 targetName "vibe_core"
 
--- a/source/vibe/internal/allocator.d
+++ b/source/vibe/internal/allocator.d
@@ -1,12 +1,12 @@
 module vibe.internal.allocator;
 
-public import std.experimental.allocator;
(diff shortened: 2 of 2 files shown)


wilzbach:
Hmm this doesn't seem to work with DMD 2.071 (and thus LDC 1.1):

image

While the non-existent `Ternary` problem and a couple of other issues with an older Phobos [can be fixed easily](https://github.com/wilzbach/stdx-allocator/commit/d06e4f2bae2eee5d380d145221ecb9cab04c90d7), one then still runs into safety issues with 2.071.2:

image

I don't really fancy wasting time looking into these and would only do so if dropping support for 2.071.1 isn't an option. [I just pushed https://github.com/wilzbach/stdx-allocator/commit/d06e4f2bae2eee5d380d145221ecb9cab04c90d7, which allows building stdx-allocator with 2.072.2.

Let me know how you would like to proceed.

s-ludwig:
I've opened #45 to drop 2.071 support.

wilzbach:
> I've opened #45 to drop 2.071 support.

Thanks! (Rebased this PR)

wilzbach:
Everything passing :tada: 

image

wilzbach:
Tagged this as v1.4.0-alpha.1, s.t. it can be used by Vibe.d

PR #46, opened 2017-12-20
Use std.parallelism.totalCPUs to implement logicalProcessorCount.

s-ludwig:
See vibe-d/vibe.d#2008.

PR #47, opened 2018-01-10
Use std.parallelism.totalCPUs instead of duplicating code

dkgroot:
This is the same as the vibe.d implementation.
BTW: This saves adding an implementation to support DragonFlyBSD (and other OS's in the future).


dkgroot:
Dropping this one, duplicate of #46

PR #49, opened 2018-01-14
avoid Buffer type which creates a huge __initZ symbol

MartinNowak:
- also see [15130 – dmd emits huge data for zero initialized struct](https://issues.dlang.org/show_bug.cgi?id=15130)

### DIFF
--- a/source/vibe/core/stream.d
+++ b/source/vibe/core/stream.d
@@ -39,12 +39,10 @@ ulong pipe(InputStream, OutputStream)(InputStream source, OutputStream sink, ulo
 	@blocking @trusted
 	if (isOutputStream!OutputStream && isInputStream!InputStream)
 {
-	import vibe.internal.allocator : theAllocator, make, dispose;
+	import vibe.internal.allocator : theAllocator, makeArray, dispose;
 
-	static struct Buffer { ubyte[64*1024] bytes = void; }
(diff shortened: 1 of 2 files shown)


source/vibe/core/stream.d
```diff
@@ -39,12 +39,10 @@ ulong pipe(InputStream, OutputStream)(InputStream source, OutputStream sink, ulo
...
 	if (isOutputStream!OutputStream && isInputStream!InputStream)
 {
-	import vibe.internal.allocator : theAllocator, make, dispose;
+	import vibe.internal.allocator : theAllocator, makeArray, dispose;
 
-	static struct Buffer { ubyte[64*1024] bytes = void; }
-	auto bufferobj = theAllocator.make!Buffer();
-	scope (exit) theAllocator.dispose(bufferobj);
-	auto buffer = bufferobj.bytes;
+	auto buffer = theAllocator.makeArray!ubyte(64*1024);
```
wilzbach: Not sure whether it's of any concern, but `makeArray` does default initialize everything.
However, there's an optimization PR pending at Phobos: https://github.com/dlang/phobos/pull/6024

MartinNowak: Uninitialized would be better, let me use raw alloc maybe.

PR #52, opened 2018-01-27
Fix incorrect return value in TCPConnection.leastSize()

nexor:
Let the method return more than just 0 or 1


wilzbach:
Ugh, AppVeyor is failing too often :/
I hope https://github.com/vibe-d/vibe-core/pull/53 helps.

PR #62, opened 2018-02-23
TCPConnection.waitForDataAsync 

FraMecca:
I have implemented your suggested changes and I am hoping that we can discuss this PR more if needed.

### DIFF
--- a/source/vibe/core/net.d
+++ b/source/vibe/core/net.d
@@ -586,33 +586,58 @@ mixin(tracer);
 		return m_context.readBuffer.length > 0;
 	}
 
-	WaitForDataAsyncStatus waitForDataAsync(void delegate(bool) @safe read_callback, Duration timeout = Duration.max)
+	/** Waits asynchronously for new data to arrive.
+
+		This function can be used to detach the `TCPConnection` from a
+		running task while waiting for data, so that the associated memory
+		resources are available for other operations.
+
+		Note that `read_ready_callback` may be called from outside of a
+		task, so no blocking operations may be performed. Instead, an existing
+		task should be notified, or a new one started with `runTask`.
+
+		Params:
+			read_ready_callback = A callback taking a `bool` parameter that
+				signals the read-readiness of the connection
+			timeout = Optional timeout to limit the maximum wait time
+
+		Returns:
+			If the read readiness can be determined immediately, it will be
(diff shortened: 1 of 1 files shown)


source/vibe/core/net.d
```diff
@@ -586,6 +586,41 @@ mixin(tracer);
...
+		if (timeout <= 0.seconds) {
+			auto rs = waitForData(0.seconds);
+			auto mode = rs ? WaitForDataAsyncStatus.dataAvailable : WaitForDataAsyncStatus.noMoreData;
+			return mode;
+		}
+
+		auto tm = setTimer(timeout, { eventDriver.sockets.cancelRead(m_socket);
+				runTask(read_callback, false); });
+		eventDriver.sockets.read(m_socket, m_context.readBuffer.peekDst(), IOMode.once,
+				(sock, st, nb) { tm.stop(); if(st != IOStatus.ok) runTask(read_callback, false);
```
s-ludwig: The two lines from above
```
assert(m_context.readBuffer.length == 0);
m_context.readBuffer.putN(nbytes);``
```

should be executed right after the `tm.stop()`, so that the read data is actually committed to the read buffer. The `if` condition would then ideally be `m_context.readBuffer.length > 0` instead of `st == IOStatus.ok`, because partial data could have been read before the error was raised (not a realistic case, since `IOMode.once` is used, but handling it this way should be slightly more robust).

source/vibe/core/net.d
```diff
@@ -586,6 +586,41 @@ mixin(tracer);
 		return m_context.readBuffer.length > 0;
 	}
 
+	enum WaitForDataAsyncStatus {
```
s-ludwig: To reduce the amount of text to type when using this from the outside, I'd move this outside below the `TCPConnection` class.

s-ludwig:
Okay, thanks, that looks good now.

I was thinking whether the `runTask` is actually such a good idea. Since someone might already have a task running in the background and would just notify it from within the callback, a new task may not always be needed and would just be overhead. For that reason it may be the better solution to just document that no blocking operations should be performed form the callback, since it may be called outside of a task, and to replace the `runTask` calls by plain callback calls. I'll just quickly add a commit to the PR for that and merge, so that this can still get into the next release.

PR #76, opened 2018-03-22
MonoTime in loopWithTimeout, and other read tweaks.

### LINKED ISSUE #39: Monotonic clock instead of SysTime for loopWithTimeout?
https://github.com/vibe-d/vibe-core/blob/master/source/vibe/core/net.d#L675

Hello again! Do you remember any special case reasoning around the choice of timing mechanism for this function? I'm positive that a timeout violation during clock switch is not a good thing.

Do you mind if I change it to MonoTime?

Boris-Barboris:
MonoTime is more robust and does not involve sophisticated timezone
shenanigans, making it better suited for networking operations.
Fixes #39.
This commit also distinguishes read timeout exception as Exception child
ReadTimeoutException, wich is thrown from loopWithTimeout function.
Optimistic read branch for tcp socket improves performance on
granular reads.

### DIFF
--- a/source/vibe/core/net.d
+++ b/source/vibe/core/net.d
@@ -659,7 +659,7 @@ mixin(tracer);
 			m_context.readBuffer.popFrontN(n);
 			count -= n;
 			return count == 0;
-		});
+		}, ReadTimeoutException);
 	}
 
 	size_t read(scope ubyte[] dst, IOMode mode)
@@ -667,6 +667,10 @@ mixin(tracer);
 mixin(tracer);
 		import std.algorithm.comparison : min;
 		if (!dst.length) return 0;
+		if (m_context.readBuffer.length >= dst.length) {
+			m_context.readBuffer.read(dst);
+			return dst.length;
+		}
 		size_t nbytes = 0;
 		m_context.readTimeout.loopWithTimeout!((remaining) {
 			if (m_context.readBuffer.length == 0) {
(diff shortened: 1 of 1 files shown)


Boris-Barboris:
some eventcore dependency problem apparently

Boris-Barboris:
@s-ludwig can you restart the ci please?

s-ludwig:
Restarted the failing Travis-CI jobs. The AppVeyor failure also looks like it's also just caused by the registry being down (unfortunately AppVeyor only allows to restart all jobs at once instead of only the failing ones). Anyway, looks good to merge! 

BTW, do you know the reasoning behind the new `Exception` constructor signature (msg, next, file, line)? I never saw that until now and, without better knowledge, would have said that supporting the second overload should be sufficient.

Boris-Barboris:
@s-ludwig 
Imagine you want to specifically set nested exception to something else. Previously you would have to explicitly type 'file' and 'line' stuff, wich is quite boring. Now you just pass message and next.

Boris-Barboris:
"root package vibe-core reference eventcore ~>0.8.32 cannot be satisfied.
Command exited with code 2"

s-ludwig:
Yes, I can see how that can be convenient, but considering that this is a feature that is officially reserved to the built-in exception chaining mechanism, I don't see how that would warrant adding a new constructor overload to the Druntime API. But a blame just revealed that this had actually been there for a long time already, it's just not documented.

s-ludwig:
> "root package vibe-core reference eventcore ~>0.8.32 cannot be satisfied.
Command exited with code 2"

This is what I meant with the registry being down. When it fails to collect package metadata, the error message will read like that instead of an explicit "failed to read package metadata for 'eventcore' from registry at code.dlang.org: ...". I'll force merge and hope that it passes ;)

PR #77, opened 2018-03-30
fix integer overflow protection absence for LocalManualEvent

Boris-Barboris:


Boris-Barboris:
> since AFAIK D adopted the unspecified overflow behavior of C/C++

No, it's good to go in D:
https://forum.dlang.org/thread/n23bo3$qe$1@digitalmars.com

Boris-Barboris:
same thing for ManualEvent

Boris-Barboris:
@s-ludwig 
dub in it's own repertoire.

What do you think of this
"core.exception.AssertError@tests/issue-58-task-already-scheduled.d(32): Timers didn't fire within the time limit" thing, I see it spontaniously asserting on old compiler versions once or twice per CI run, but I cannot reproduce it locally. Was this thing happening before me?

wilzbach:
Yes, the history shows up a lot of other failures:

https://travis-ci.org/vibe-d/vibe-core/builds

s-ludwig:
Yeah, unfortunately the timeout behavior of epoll/kqueue somtimes is totally off on the CI machines for some reason. I tried scaling up the times, but it almost seems as if the timeout sometimes has no effect at all.

PR #79, opened 2018-04-07
Fix Task.running and Task.yield directly after runTask/yieldLock.

s-ludwig:
The status of the task was erroneously reported as not running until the task was actually executed for the first time (which only happens after the yield lock has been lifted).


Extrawurst:
@s-ludwig stumbled overt this I think - can we get a new beta-tag?

s-ludwig:
@Extrawurst will do after merging #77.

PR #90, opened 2018-09-19
fix #2211 - set socket to invalidate state when connection is closed

Kozzi11:
fix: https://github.com/vibe-d/vibe.d/issues/2211

### DIFF
--- a/source/vibe/core/net.d
+++ b/source/vibe/core/net.d
@@ -586,7 +586,7 @@ mixin(tracer);
 				break;
 			case IOStatus.ok: break;
 			case IOStatus.wouldBlock: assert(mode == IOMode.immediate); break;
-			case IOStatus.disconnected: break;
+			case IOStatus.disconnected: m_socket = StreamSocketFD.invalid; break;
 		}
 
 		return m_context.readBuffer.length > 0;
@@ -735,7 +735,7 @@ mixin(tracer);
 			default:
 				throw new Exception("Error writing data to socket.");
 			case IOStatus.ok: break;
-			case IOStatus.disconnected: break;
+			case IOStatus.disconnected: m_socket = StreamSocketFD.invalid; break;
 		}
 
 		return res[2];
--- a/tests/vibe.core.net.2211.d
+++ b/tests/vibe.core.net.2211.d
@@ -1,0 +1,37 @@
+/+ dub.sdl:
+name "test"
+description "TCP disconnect task issue"
+dependency "vibe-core" path="../"
++/
+module test;
+
+import vibe.core.core;
+import vibe.core.net;
+import core.time : msecs;
+import std.string : representation;
+
+void main()
+{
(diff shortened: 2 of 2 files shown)


Kozzi11:
@s-ludwig Is there anything wrong with this pull request?

s-ludwig:
I'm not totally sure about the concrete case, but in general there is an asymmetry between `.empty` and `.connected` (latter not being an ideal naming choice, unfortunately), where `.empty` determines the closed state of the read direction and `.connected` determines the same for the write direction. Only `.empty && !.connected` means that the connection has been fully closed. This behavior is mentioned in the documentation of [`connected`](http://vibed.org/api/vibe.core.stream/ConnectionStream.connected).

I thought that there was a test case for this, but looking through the tests it appears that I was wrong. I'll try to add one in the near future to get this pinned down properly.

Kozzi11:
My problem with empty is this:

    For connection oriented streams, this function will block until either new data arrives or the connection got closed.

I need to do something like this:

    while(conn.connected) // can not use empty because it will block
    {
            if (conn.waitForData(dur!"msecs"(10)))
            {
                // read data and do something with data
            }
            else
            {
                // do something else
            }
    }
    
I can not find how to find if waitForData timeouted or connection has been closed

Marenz:
> his behavior is mentioned in the documentation of connected.

Hu, that is very unintuitiv. And sure enough, I was checking for `connected` in my reader task. That might explains a few of my bugs.

Might I suggest a rename? I am not sure into what yet. Maybe.. `write_closed` and `read_closed` or so..

s-ludwig:
> I can not find how to find if waitForData timeouted or connection has been closed

That's indeed a good point. The only way I see right now is to call `waitForDataAsync` with a non-zero timeout and see if it returns `WaitForDataAsyncStatus.noMoreData` after `waitForData` already returned `false` before. There should probably just be a `@property ConnectionState connectionState()` to be able to determine the full state.

> Might I suggest a rename? I am not sure into what yet. Maybe.. write_closed and read_closed or so..

This is definitely something to think about as this is indeed a recurring source of misunderstandings. I would suggest to just change the name of `connected` to something that makes it clear that it belongs to `write` and keep `empty` for the read side, as everything else would simply cause too much breakage.

s-ludwig:
There is now `waitForDataEx`, which returns one of `dataAvailable`, `noMoreData` or `timeout`, so the loop would look like this:
```
while (true) {
    auto res = conn.waitForDataEx(10.msecs);
    if (res == WaitForDataStatus.noMoreData)
        break;

    if (res == WaitForDataStatus.dataAvailable) {
            // read data
    } else {
            // do something else
    }
}
```

s-ludwig:
Any objection to close this?

Kozzi11:
No, `waitForDataEx` is good enought. Thank you

PR #95, opened 2018-10-23
Avoid starting an explicit event loop in worker threads.

s-ludwig:
This slightly reduces the overhead of task pool threads and simplifies stack traces in case of errors.


wilzbach:
Though it doesn't seem to work with older D versions:

```
core.exception.AssertError@tests/issue-58-task-already-scheduled.d(32): Timers didn't fire within the time limit
----------------
??:? [0x4b31ee]
```

s-ludwig:
I hope that the CI servers just had a bad day, because in theory no worker threads should be involved in those tests. I'll restart to see what happens.

PR #108, opened 2018-12-07
UDP connect wrongly passes port as addr family to resolveHost

Marenz:


Marenz:
Update: 
- Removed trailing whitespace

PR #116, opened 2019-01-08
release ref on failing connectTCP

### LINKED ISSUE #115: connectTCP failure leaking eventcore driver handles
```
Running ./test --vvvv
[2019-01-08 22:17:44.844 830AA14F(----) trc] create driver core
[2019-01-08 22:17:44.844 main(----) dbg] Thread exit main (index 0) (main=true)
[2019-01-08 22:17:44.844 main(----) dia] Main thread exiting
Warning (thread: main): leaking eventcore driver because there are still active handles
   FD 6 (streamSocket)
   FD 7 (streamSocket)
   FD 8 (streamSocket)
   FD 9 (streamSocket)
   FD 10 (streamSocket)
   FD 11 (streamSocket)
   FD 12 (streamSocket)
   FD 13 (streamSocket)
   FD 14 (streamSocket)
   FD 15 (streamSocket)
   FD 16 (streamSocket)
   FD 17 (streamSocket)
   FD 18 (streamSocket)
   FD 19 (streamSocket)
   FD 20 (streamSocket)
   FD 21 (streamSocket)
   FD 22 (streamSocket)
   FD 23 (streamSocket)
   FD 24 (streamSocket)
   FD 25 (streamSocket)
Warning (thread: main): leaking eventcore driver because there are still active handles
   FD 6 (streamSocket)
   FD 7 (streamSocket)
   FD 8 (streamSocket)
   FD 9 (streamSocket)
   FD 10 (streamSocket)
   FD 11 (streamSocket)
   FD 12 (streamSocket)
   FD 13 (streamSocket)
   FD 14 (streamSocket)
   FD 15 (streamSocket)
   FD 16 (streamSocket)
   FD 17 (streamSocket)
   FD 18 (streamSocket)
   FD 19 (streamSocket)
   FD 20 (streamSocket)
   FD 21 (streamSocket)
   FD 22 (streamSocket)
   FD 23 (streamSocket)
   FD 24 (streamSocket)
   FD 25 (streamSocket)
```

Reproduce:
```d
#!/usr/bin/env dub
/+ dub.sdl:
name "test"
dependency "vibe-core" version="~>1.4.6"
+/

import core.time;

import vibe.core.core;
import vibe.core.net;

void main()
{
	foreach (_; 0 .. 20)
	{
		TCPConnection conn;
		try
		{
			conn = connectTCP("127.0.0.1", 16565); // any unused port
		}
		catch (Exception e)
		{
		}
	}
}
```

Also happens when using the overload taking a NetworkAddress

WebFreak001:
fix #115 

Unsure how this affects windows, ~~but if the auto tests are also run on windows it should find any issue there.~~ please test manually

status is refused in this case and it's not handled by the existing cases cancelling a connection, so the reference to the socket still exists and needs to be released by the caller here.

### DIFF
--- a/source/vibe/core/net.d
+++ b/source/vibe/core/net.d
@@ -218,8 +218,10 @@ TCPConnection connectTCP(NetworkAddress addr, NetworkAddress bind_address = anyA
 			": timeout");
 
 		if (status != ConnectStatus.connected) {
-			if (sock != SocketFD.invalid)
-				assert(!eventDriver.sockets.releaseRef(sock));
+			if (sock != SocketFD.invalid) {
+				bool refsLeft = eventDriver.sockets.releaseRef(sock);
+				assert(!refsLeft);
+			}
 
 			enforce(false, "Failed to connect to "~addr.toString()~": "~status.to!string);
 			assert(false);
--- a/source/vibe/core/net.d
+++ b/source/vibe/core/net.d
@@ -216,7 +216,14 @@ TCPConnection connectTCP(NetworkAddress addr, NetworkAddress bind_address = anyA
 
 		enforce(!cancelled, "Failed to connect to " ~ addr.toString() ~
 			": timeout");
-		enforce(status == ConnectStatus.connected, "Failed to connect to "~addr.toString()~": "~status.to!string);
+
+		if (status != ConnectStatus.connected) {
+			if (sock != SocketFD.invalid)
+				assert(!eventDriver.sockets.releaseRef(sock));
+
+			enforce(false, "Failed to connect to "~addr.toString()~": "~status.to!string);
+			assert(false);
+		}
 
 		return TCPConnection(sock, uaddr);
 	} ();
--- a/tests/issue-115-connect-fail-leak.d
+++ b/tests/issue-115-connect-fail-leak.d
@@ -1,0 +1,22 @@
+/+ dub.sdl:
+	name "test"
+	dependency "vibe-core" path=".."
++/
+module test;
+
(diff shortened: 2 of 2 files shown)


s-ludwig:
It looks like this is an issue in eventcore. All other error codes result in the socket being closed, but `ConnectionStatus.refused` keeps it open (direct reason for the difference is that all other codes are generated immediately, while `refused` is an event loop event). I'll add a call to `releaseRef` right after the user callback has been invoked, so that the socket handle is still valid during the callback.

s-ludwig:
Upon taking a closed look, the behavior in eventcore makes sense, because otherwise the handle already returned by `connectStream` would be silently invalidated. So fixing it here is the right thing to do after all.

It would be nice to get the test case to actually fail. The only way I see right now is to count the number of open files at the start and at the end of the test case (e.g. [using `getrlimit`](https://stackoverflow.com/questions/7976769/getting-count-of-current-used-file-descriptors-from-c-code)). I'd look into the Windows version of this.

WebFreak001:
Can't we make the "get leaked handles" message as a public function in eventcore?

s-ludwig:
Doing that properly would be a bit more involved, because the leaked handles may be scattered over multiple threads/drivers (in TLS-memory), so the first step would be a global registry for driver instances and thread-safe way to query leaked handles of foreign threads. The `getrlimit` variant would be a oneliner using `iota`/`count` on the other hand.

s-ludwig:
I'll tackle the test additions in a separate branch.

PR #128, opened 2019-01-22
remove refsLeft assert

WebFreak001:
It seems when running in a vibe-d app listening to HTTP requests that the assert fails. Even though I would like there to be a check, closing the vibe.d app afterwards shows no leaked handles so I think it's fine, and definitely better than not starting anymore at all.

### DIFF
--- a/source/vibe/core/net.d
+++ b/source/vibe/core/net.d
@@ -217,10 +217,8 @@ TCPConnection connectTCP(NetworkAddress addr, NetworkAddress bind_address = anyA
 			": timeout");
 
 		if (status != ConnectStatus.connected) {
-			if (sock != SocketFD.invalid) {
-				bool refsLeft = eventDriver.sockets.releaseRef(sock);
-				assert(!refsLeft);
-			}
+			if (sock != SocketFD.invalid)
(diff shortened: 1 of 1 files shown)


s-ludwig:
Would be nice to have a reproduction case for this. Is it enough to start a HTTP server and then attempt a failing outgoing connection?

WebFreak001:
yep this triggers it:
```d
import vibe.vibe;

void main()
{
	runTask({
		auto conn = connectTCP("127.0.0.1", 16516);
		assert(false);
	});

	auto set = new HTTPServerSettings();
	set.port = 8080;
	listenHTTP(set, &index);

	runEventLoop();
}

void index(scope HTTPServerRequest req, scope HTTPServerResponse res)
{
}
```

s-ludwig:
Okay, thanks, it even triggers without the HTTP server for me: https://run.dlang.io/is/x97pL3

The reason is that within a task, the task will be resumed from within the eventcore callback, during which there is still an extra reference present to keep the FD valid.

On Windows it actually works, because it has different semantics (it only calls the callback if there is at least one reference left). I need to think about this some more before unifying them.

PR #130, opened 2019-01-23
Channel improvements

s-ludwig:
The branch is based on #129, which should be merged first.

- Makes the API more `@safe` and `nothrow`
- Adds `Channel.bufferFill`

### DIFF
--- a/source/vibe/core/channel.d
+++ b/source/vibe/core/channel.d
@@ -54,6 +54,14 @@ struct Channel(T, size_t buffer_size = 100) {
 	/// ditto
 	@property bool empty() shared { return m_impl.empty; }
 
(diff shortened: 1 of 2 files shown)


source/vibe/core/channel.d
```diff
@@ -18,6 +18,8 @@ import core.sync.mutex;
 // TODO: implement a multi-channel wait, e.g.
 // TaggedAlgebraic!(...) consumeAny(ch1, ch2, ch3); - requires a waitOnMultipleConditions function
 
+// NOTE: not using synchronized (m_mutex) because iot is not nothrow
```
l-kramer: *it

PR #149, opened 2019-04-16
Fix #146.

### LINKED ISSUE #146: on macOS: std.json.JSONException@std/json.d(155): JSONValue is not a string
in dub.json:

```
{
	"name" : "d_test007",
	"description" : "Hello World - A minimal DUB bundle.",
	"dependencies" : {
		"vibe-core": "1.6.2"
	}
}
```

in src/app.d:

```
import vibe.core.log;

void main()
{
}

```
Compile and run it, output:
**std.json.JSONException@std/json.d(155): JSONValue is not a string**

DMD: 2.084.0, 2.085.0, 2.085.1
OS: MacOS 10.14.3

shove70:
Fix #146: on macOS: std.json.JSONException@std/json.d(155): JSONValue is not a string

### DIFF
--- a/source/vibe/core/args.d
+++ b/source/vibe/core/args.d
@@ -304,7 +304,7 @@ private T fromValue(T)(in JSONValue val)
 	static if (is(T == bool)) return val.type == JSON_TYPE.TRUE;
 	else static if (is(T : long)) return val.integer.to!T;
 	else static if (is(T : double)) return val.floating.to!T;
-	else static if (is(T == string)) return val.str;
+	else static if (is(T == string)) return (val.type == JSON_TYPE.STRING) ? val.str : val.to!T;
 	else static assert(false);
 
 }
--- a/source/vibe/core/args.d
+++ b/source/vibe/core/args.d
@@ -304,7 +304,7 @@ private T fromValue(T)(in JSONValue val)
 	static if (is(T == bool)) return val.type == JSON_TYPE.TRUE;
 	else static if (is(T : long)) return val.integer.to!T;
 	else static if (is(T : double)) return val.floating.to!T;
-	else static if (is(T == string)) return (val.type == JSON_TYPE.STRING) ? val.str : val.to!T;
+	else static if (is(T == string)) return val.str;
 	else static assert(false);
 
 }
--- a/source/vibe/core/core.d
+++ b/source/vibe/core/core.d
@@ -1434,8 +1434,18 @@ shared static this()
 
 	version (VibeNoDefaultArgs) {}
 	else {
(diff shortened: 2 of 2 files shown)


s-ludwig:
Did you have a look at your `/etc/vibe/vibe.conf` or `~/vibe.conf` (if any of those exist) to see which field causes the exception?

What I'm thinking is that the most appropriate approach to deal with this is to improve the generated error message to somthing like `"Malformed configuration field 'xxx' in /etc/vibe/vibe.conf: JSONValue is not a string"`. Just converting to string could lead to another exception later on, which may be even harder to track down to the root cause.

shove70:
$ cat /etc/vibe/vibe.conf
```
{
	"uid": 70,
	"gid": 70
}
```
"uid" and "gid" are int types.

./source/vibe/core/core.d:   (line: 1435-1439)
```
	version (VibeNoDefaultArgs) {}
	else {
		readOption("uid|user", &s_privilegeLoweringUserName, "Sets the user name or id used for privilege lowering.");
		readOption("gid|group", &s_privilegeLoweringGroupName, "Sets the group name or id used for privilege lowering.");
	}
```

s_privilegeLoweringUserName and s_privilegeLoweringGroupName are string types, So the template parameter that calls the template readOption is string type.

So, the more complicated modification is to change here.
Where do you think it should be revised?

shove70:
Like this?

```
	version (VibeNoDefaultArgs) {}
	else {
		readOption("user", &s_privilegeLoweringUserName, "Sets the user name or id used for privilege lowering.");
		readOption("group", &s_privilegeLoweringGroupName, "Sets the group name or id used for privilege lowering.");

		if (s_privilegeLoweringUserName.empty || s_privilegeLoweringGroupName.empty)
		{
			int uid, gid;
			readOption("uid", &uid, "Sets the user name or id used for privilege lowering.");
			readOption("gid", &gid, "Sets the group name or id used for privilege lowering.");

			s_privilegeLoweringUserName = uid.to!string;
			s_privilegeLoweringGroupName = gid.to!string;
		}
	}
```

shove70:
In the vibe.d project, the vibe.conf file generated by the shell file ./setup-mac.sh,

If quotation marks are added to the values of "uid" and "gid", this problem can also be avoided. Which solution do you think is more appropriate?

```
echo -e '{
	"uid": '$USER_ID',
	"gid": '$GROUP_ID'
}' >$CONFIG_FILE
```

Modified to:

```
echo -e '{
	"uid": "'$USER_ID'",
	"gid": "'$GROUP_ID'"
}' >$CONFIG_FILE
```

PR #150, opened 2019-04-16
Add way to disable port and address reuse

rracariu:
This integrates with recent event core `reuseAddress` flag.
It allows one to selectively disable those flags.


s-ludwig:
Thanks! I totally forgot that vibe.d has its own set of flags...

We should do the same change with `defaults` and `none` for `TCPListenOptions` as with `StreamListenOptions`, though.

rracariu:
Fixed as suggested.

PR #153, opened 2019-04-26
Use timeout with `leastSize` on TCPConnection

tchaloupka:
I'm not exactly sure about this, but I guess that if one sets the read timeout on the `TCPConnection` it should be used in this blocking call too so it won't block forever.

### DIFF
--- a/source/vibe/core/net.d
+++ b/source/vibe/core/net.d
@@ -593,7 +593,7 @@ mixin(tracer);
 			}
 		);
 
-		asyncAwaitAny!(true, waiter)(timeout);
+		enforce!ReadTimeoutException(asyncAwaitAny!(true, waiter)(timeout), "Read operation timed out");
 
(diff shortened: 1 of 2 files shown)


s-ludwig:
Agreed about the timeout. It should throw an exception in that case, though, because otherwise `leastSize` would signal a connection close with `0` bytes returned.

tchaloupka:
I've modified the `asyncAwaitAny` with timeout parameter to return if `Waitable` completed in time or not and then used it in `waitForData`.

Let me know if it's ok this way.

s-ludwig:
The main problem is that `waitForData` may not throw in case of a timeout, I've opened a PR with an alternative implementation that I had lying around for a while: #159 With this, `leastSize` simply has to `enforce(ret != WaitForDataStatus.timeout)`.

s-ludwig:
Okay, #159 is in now.

PR #154, opened 2019-05-05
Implement vibe.core.process for subprocess handling similar to std.process

BenjaminSchaaf:
I've tried to keep most of the features of std.process, if there's anything big I missed let me know.

The tests are simple ports from [supervised](https://github.com/BenjaminSchaaf/supervised). They don't have full coverage, but the core should be solid.

### DIFF
--- a/source/vibe/core/process.d
+++ b/source/vibe/core/process.d
@@ -30,15 +30,15 @@ import vibe.core.internal.release;
     started from anywhere including external libraries or std.process.
 
     Params:
-        pid = A Pid or OS process handle
+        pid = A Pid or OS process id
 */
-Process registerProcess(Pid pid)
+Process adoptProcessID(Pid pid)
 @trusted {
-    return registerProcess(pid.processID);
+    return adoptProcessID(pid.processID);
 }
 
 /// ditto
-Process registerProcess(int pid)
+Process adoptProcessID(int pid)
 {
     return Process(eventDriver.processes.adopt(pid));
 }
--- a/source/vibe/core/process.d
+++ b/source/vibe/core/process.d
@@ -1,0 +1,752 @@
+/**
+    Functions and structures for dealing with subprocesses and pipes.
+
+    This module is modeled after std.process, but provides a fiber-aware
+    alternative to it. All blocking operations will yield the calling fiber
+    instead of blocking it.
+*/
+module vibe.core.process;
+
+public import std.process : Pid, Redirect;
+static import std.process;
+
+import core.time;
+import std.array;
+import std.typecons;
+import std.exception : enforce;
+import std.algorithm;
+import eventcore.core;
+import vibe.core.path;
+import vibe.core.log;
+import vibe.core.stream;
+import vibe.internal.async;
+import vibe.internal.array : BatchBuffer;
+import vibe.core.internal.release;
+
+@safe:
+
+/**
+    Register a process with vibe for fibre-aware handling. This process can be
+    started from anywhere including external libraries or std.process.
+
+    Params:
+        pid = A Pid or OS process handle
+*/
+Process registerProcess(Pid pid)
+@trusted {
+    return registerProcess(pid.osHandle);
+}
+
+/// ditto
+Process registerProcess(int pid)
+{
+    return Process(eventDriver.processes.adopt(pid));
+}
+
+/**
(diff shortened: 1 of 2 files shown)


source/vibe/core/process.d
```diff
@@ -0,0 +1,758 @@
...
+    Collect the string output of a stream in a blocking fashion.
+
+    Params:
+        stream = The input stream to collect from.
+        nbytes = The maximum number of bytes to collect.
+
+    Returns:
+        The collected data from the stream as a string.
+*/
+string collectOutput(InputStream)(InputStream stream, size_t nbytes = size_t.max)
```
s-ludwig: In general, it probably makes sense to eventually move `readAll(UTF8)` to `vibe.core.stream` instead of `vibe.stream.operations`, since it is such basic functionality, similar to `pipe`.

Until then, since this is more or less the same, this one should have a doc comment `/// private` to remove it from the documentation (the current description can stay as a normal comment that precedes the doc comment).

source/vibe/core/process.d
```diff
@@ -0,0 +1,758 @@
...
+@safe:
+
+/**
+    Register a process with vibe for fibre-aware handling. This process can be
+    started from anywhere including external libraries or std.process.
+
+    Params:
+        pid = A Pid or OS process handle
+*/
+Process registerProcess(Pid pid)
```
s-ludwig: My initial thought for the name was `openProcess(by(P)ID)`, probably because that is what the Windows API uses. Register process may sound a bit like it would statically register the process somewhere instead of just returning a handle. What do you think?

tests/vibe.core.process.d
```diff
@@ -0,0 +1,206 @@
...
+}
+
+void testLineEndings()
+{
+    auto program = q{
+        write("linux\n");
+        write("os9\r");
+        write("win\r\n");
+    };
+    auto res = execute(["rdmd", "--eval", program]);
```
s-ludwig: The eval argument fails for DMD < 2.080.0, I'd suggest to simply `static if (__VERSION__ >= 2080)` this whole test file instead of finding a workaround, as there is nothing in the test that should be compiler version specific (e.g. templated code that is only instantiated by the test).

source/vibe/core/process.d
```diff
@@ -0,0 +1,758 @@
...
+    unittest {
+        Process p;
+
+        assert(!p);
+    }
+
+    /**
+        An operating system handle to the process.
+    */
+    @property int osHandle() const nothrow @nogc { return cast(int)m_pid; }
```
s-ludwig: If there is nothing that speaks against it, I'd suggest to simply stick to `pid` as the name, as handle is slightly confusing (Windows uses `HANDLE` to access processes, in addition to the usual PID).

BenjaminSchaaf:
Thanks for the review @s-ludwig, it would also be good to merge this: https://github.com/vibe-d/eventcore/pull/110, as it causes some bugs in this code

source/vibe/core/process.d
BenjaminSchaaf: Maybe `adoptProcess` would be better and more in-line with eventcore?

source/vibe/core/process.d
BenjaminSchaaf: Ah, I wasn't aware windows had both PID and a HANDLE. `pid` makes a lot more sense then :)

source/vibe/core/process.d
s-ludwig: Thought about that, too, but dismissed it at the time, because it also sounded a bit like it would do something with the process as opposed to just creating a "handle" for it. But apart from eventcore using it, another argument for "adopt" would be consistency with a possible analoguous socket API - `openSocket` would be rather misleading there.

I think the mentioned issue can be fixed quite easily, though, by including "ID" in the name (i.e. `adoptProcessID` or `adoptPID`), so "adopt" sounds fine after all.

BenjaminSchaaf: Good idea, I'll make the change

s-ludwig:
Thanks again! Looks good to merge now. I'll add tag a new alpha tag afterwards (or maybe even beta).

BenjaminSchaaf:
Thanks for merging this! Looking forward to using this in my projects :)

PR #171, opened 2019-08-20
Clean up std.json deprecation, remove DUB hack in CI

Geod24:
CC @s-ludwig : I was thinking of doing the same to Vibe.d, as the CI nowadays takes half a day for a single PR.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,8 +41,7 @@ branches:
     - master
 
 before_install:
-  # Use the dub-updating fork of the installer script until https://github.com/dlang/installer/pull/301 is merged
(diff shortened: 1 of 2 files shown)


s-ludwig:
I've reduced the number of tested compilers to the latest four + the oldest support version in #166. The past has shown that there are always people who rely on certain compiler versions for a longer time, so I'd like to keep support for the old versions as long as possible or desirable w.r.t. to new language features.

Not sure about the usefulness of the list of former supported compiler versions, but another idea would be to generally simplify the list to "DMD 2.076.0 - 2.087.1" and then just keep the list of old versions, too.

Geod24:
@s-ludwig : Updated

s-ludwig:
Thanks! I've fixed a compile error and toggled auto-merge.

PR #188, opened 2020-01-06
Replace 'body' with 'do'

Geod24:


Geod24:
The failures look spurious but I have no way to retrigger the CI

Geod24:
Rebased on master, added a fix for the badge in the README, let's see if this passes now...

s-ludwig:
Looking good!

Geod24:
If you hit the retry button a dozen times, it ultimately does it seems...

s-ludwig:
Okay, that's a lot. Did you see whether it was always the tests/std.concurrency.d test, or also sometimes the dirwatcher.d one? Both used to be failing about equally frequent, but in the previous version of this PR it seemed like it was always the concurrency one. But it might just have been the time of day/job load.

s-ludwig:
Hm okay, so one of the jobs already stops with an assertion for `dub test` and the other one failed the concurrency test. I'll restart again.

Geod24:
> Did you see whether it was always the tests/std.concurrency.d test, or also sometimes the dirwatcher.d one?

I've seen both of them a lot, but the concurrency one seemed to never go away.
In any case, I am flabbergasted that such a simple change could trigger a consistent failure, and kinda afraid, too.

s-ludwig:
Hm, so it does seem to have something to do with the CI load, but that dub test assertion is still very worrying. But, although I currently have no idea why, the reason why this PR in particular fails so much appears to be that LDC 1.15.0 and 1.18.0 are not tested at all on master/macOS so far: https://github.com/vibe-d/vibe-core/runs/594339328

The latest PR didn't even run on any LDC version for macOS: https://github.com/vibe-d/vibe-core/runs/592170278 - This almost looks like a probabilistic way of reducing the macOS worker load.

s-ludwig:
Anyway, all of this makes it likely enough that this is indeed a false alarm, so I'm going to merge this before looking into this more.

Geod24:
Woooohhooooo, thanks!

By the way @s-ludwig I recently took a liking in Github Actions, and would recommend anyone to look into them for their CI pipelines. Simpler enough to do cross-platform dev, and blazingly fast. I plan to add support for them to `dlang/ci` so that we can transition a few projects.

PR #192, opened 2020-02-17
Update list of supported compilers

Geod24:
Since LDC v1.9.0 and v1.14.0 are consistently failing, remove them from the CI.
It's not my preferred solution to remove a failing test, but no one seems interested in fixing this (master has been red for over 2 months), and it seems to be an LDC specific issue (so not a language thing).

See commit-per-commit for an explanation of the whats and whys.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -68,7 +68,5 @@ before_install:
 
 script: ./travis-ci.sh
 
-sudo: false
-
 after_success:
  - bash <(curl -s https://codecov.io/bash)
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,8 +14,6 @@ d:
   - ldc-1.17.0
   - ldc-1.16.0
   - ldc-1.15.0
-  - ldc-1.14.0
-  - ldc-1.9.0
   - dmd-2.087.1
   - dmd-2.086.1
   - dmd-2.085.1
@@ -44,10 +42,6 @@ matrix:
       - d: ldc-1.16.0
         os: osx
       - d: ldc-1.15.0
-        os: osx
-      - d: ldc-1.14.0
-        os: osx
-      - d: ldc-1.9.0
         os: osx
   include:
     - d: dmd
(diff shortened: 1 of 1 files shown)


s-ludwig:
Hm, am I looking at the right build here?
https://travis-ci.org/vibe-d/vibe-core/builds/642558400
https://ci.appveyor.com/project/s-ludwig/vibe-core/branch/master

Both master builds are green for LDC 1.9.0/1.14.0, AFAICS. There have been two failures for DMD, though. I have restarted them to see whether they are flaky.

s-ludwig:
Ah, now I see, are you are looking at #188? I've restarted the failing builds, but it seems like the changes actually do somehow trigger them. I haven't seen those on master and the restarted builds fail in the same way.

Geod24:
Yes I was under the impression that it triggered on master as well, but you're right it doesn't, and the change do trigger them. There is no consistent failure, only a somewhat flakey dirwatcher test.

Now looking again at #188 I can't really figure out *why* this set of change would trigger *this* bug... But that's a separate topic. Should I just reinstante 1.4.0 and 1.14.0 LDC ?

.travis.yml
```diff
@@ -9,18 +9,16 @@ d:
   # this way the overall test time gets cut down (GDC/LDC are a lot
   # slower tham DMD, so they should be started early), while still
   # catching most DMD version related build failures early
-  - dmd-2.088.0
+  - dmd-2.090.1
   - dmd-2.078.3
-  - ldc-1.17.0
-  - ldc-1.16.0
+  - ldc-1.20.0
```
s-ludwig, line 14:
```suggestion
  - ldc-1.20.1
```

(for attempting to fix the CI error and to retrigger with the changed CI configuration)

Geod24:
Updated with LDC 1.20.1 and DMD 2.091.0

Geod24:
Travis is green, but I've seen spurious failures

s-ludwig:
I have restarted them, they happen on the macOS VMs due to their reliance on timing the huge spikes that occur there. One day I'll fix them with proper synchronization primitives.

Geod24:
Since I am developing on Mac OSX, I'd love to see that day :D
Thanks!

PR #198, opened 2020-03-16
Trivial improvements

s-ludwig:


Geod24:
> Use normal main() and avoid socket leak in test.

Is it a good time to mention https://github.com/vibe-d/vibe.d/issues/2370 ? :D

PR #212, opened 2020-07-24
Improve logging routines by using > 2.079 feature

Geod24:
See each commit for the description of what they do.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,13 +9,13 @@ d:
   # this way the overall test time gets cut down (GDC/LDC are a lot
   # slower tham DMD, so they should be started early), while still
   # catching most DMD version related build failures early
-  - dmd-2.092.0,dub
-  - dmd-2.078.3,dub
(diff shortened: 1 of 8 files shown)


s-ludwig:
Just noting that the log API changes are technically breaking changes that require a major version bump, even if nobody is likely to be affected by it.

Geod24:
Right. I wasn't quite sure how to proceed here.

1) Obviously making a new major for this was overkill;
2) Even if I went for a new major, what policy should be followed ? Should all new developments go to the `v1.x.x` release, or directly to the new master ?
3) A new major in a library that follow semver entails a new major in all downstream libraries. That means more work on Vibe.d & anything that depends on them.

So instead I checked that this did not affect Vibe.d (and looked for other users).
How would you prefer we address point 2 ? By the way the motivation for this PR was not merely to clean up things, but we were having issues debugging our server because some Exceptions trace would not show even at maximum trace level.

PR #228, opened 2020-09-18
Fix InterfaceProxy to work with null values.

s-ludwig:
Previously caused crashes and opCast!bool to return true for null instances.

### DIFF
--- a/source/vibe/internal/interfaceproxy.d
+++ b/source/vibe/internal/interfaceproxy.d
@@ -97,14 +97,24 @@ struct InterfaceProxy(I) if (is(I == interface)) {
 	}
 
 	this(O)(O object) @trusted
+		if (is(O == struct) || is(O == class) || is(O == interface))
 	{
 		static assert(O.sizeof % m_value[0].sizeof == 0, "Sizeof object ("~O.stringof~") must be a multiple of a pointer size.");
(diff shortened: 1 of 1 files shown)


wilzbach (approved):

wilzbach:
We should probably replace those flaky AppVeyor tests with GitHub actions.

s-ludwig:
> We should probably replace those flaky AppVeyor tests with GitHub actions.

Yeah, would be nice to unify the CI. I haven't looked into GitHub actions so far, though (maybe I should have done that instead of migrating to appveyor.com).

wilzbach:
> > We should probably replace those flaky AppVeyor tests with GitHub actions.
> 
> Yeah, would be nice to unify the CI. I haven't looked into GitHub actions so far, though (maybe I should have done that instead of migrating to appveyor.com).

If you enable GitHub actions for this repo, I can play with it on https://github.com/vibe-d/vibe-core/pull/229.

s-ludwig:
Actions are enabled, could it be that the workflows name "test.yml" requires additional configuration before it gets executed?

wilzbach:
Hmm, they run here: https://github.com/wilzbach/vibe-core/actions/runs/260953798
I think I need to open the PR from a branch of this repo -> https://github.com/vibe-d/vibe-core/pull/230

PR #230, opened 2020-09-18
Add initial GitHub actions

wilzbach:
See https://github.com/vibe-d/vibe-core/pull/228#issuecomment-694762397

Just to give an initial start.

Details:
- https://github.com/dlang-community/setup-dlang
- Advanced example: https://github.com/dlang/dmd/blob/master/.github/workflows/runnable_cxx.yml


s-ludwig:
Hm, is it normal that some of the jobs are still pending after hours?

wilzbach:
> Hm, is it normal that some of the jobs are still pending after hours?

Not really. Though, it seems to have worked flawless on the rerun I triggered.

I have now added all jobs that are run on AppVeyor, s.t. we could already remove AppVeyor.
We probably need to see how good the through-put is before replacing Travis as there's a lot of long-running jobs there.

wilzbach:
We'll need to update the README shield (see e.g. https://github.com/badges/shields/issues/2574#issuecomment-559344943), but I think this is only available once it is in the main branch.

s-ludwig:
It says 3.x minutes, which would be awesome, although AppVeyor also did a release build. I have disabled the AppVeyor check now and made the GitHub ones required instead.

Issue #205, opened 2020-04-10
runEventLoop fails to exit even when receiving signal


schveiguy:
On my Linux system (linux mint 19), I install a vanilla vibe.d server. Run it, and kill it with ctrl-c. I get a message that the server is shutting down, and get back the command prompt.

But the server didn't shut down, it's still running. And holding onto the listening socket. Turns out it doesn't exit the event loop. This is new since 1.9.0. 1.8.1 does not have this problem.

Attached is a sample log from my system.
[badvibe.txt](https://github.com/vibe-d/vibe-core/files/4462864/badvibe.txt)

Kozzi11:
You can fix it for now by define VibeHighEventPriority version

jamesragray:
How do you define VibeHighEventPriority version?

Geod24:
In dub.json:
```
"versions": [ "VibeHighEventPriority" ]
```

schveiguy:
@Kozzi11 are there any negative effects to expect from that version definition? I'm going to try it, as I'm running into issues with 1.8.1.

Kozzi11:
@schveiguy  Everything works ok for me

Robert-M-Muench:
I have the same problem on OSX.

s-ludwig:
With eventcore 0.9.11 this should now be  fixed (works for me at least).

macOS also works when building with the `kqueue` configuration, but still fails to quit with `cfrunloop`, because AFAICS there is no way to make `CFRunLoopRunInMode` (or some other equivalent API) exit after a signal. I'm also not aware of something similar to signalfd on macOS, so I'm not sure whether this can be made to work at all (disclaimer: I'm not a mac expert).

schveiguy:
I can confirm, my test vanilla vibe server works when I update the dependency! If I create a new vibe-d server, also works.

Thanks!

PR #237, opened 2020-11-25
fix issue #205 Resume task immediately if there is no more task in queue

Kozzi11:


s-ludwig:
I looked through the code really closely to see where the issue might be, but didn't find an explanation why the current logic should be wrong. However, when I examined the debug output, it finally became clear that this is actually an effect specific to the Ctrl+C signal handling code.

What happens is that the process is in an infinite wait state when Ctrl+C is pressed, causing the interrupt signal handler to be called. This in turn fires a `ManualEvent`, which in turn tries to wake up the task in question. The wait call (`select`/`epoll_wait`/`kevent`) will then return with `-1` (`EINTR`), which currently causes it to return `false` aka "no events received". The event processing logic in eventcore will then simply call it again until the timeout is actually reached (never), instead of giving the vibe-core scheduler a chance to start the task.

The fix for this is simply to return `true`, pretending/assuming that a kind of event has occurred. I'll open a PR in eventcore for that.

Thanks a lot for finding this workaround, although it may not be a full fix, it finally pushed me in the right direction!

s-ludwig:
PR: https://github.com/vibe-d/eventcore/pull/163

Kozzi11:
Yes, I have find out it too, when I have enabled all debugs I have seen it will end up in task.d:879:

PR #242, opened 2021-01-12
Rework the listDirectory implementation

s-ludwig:
- Directly uses OS facilities instead of Phobos to avoid string processing overhead and to enable fast skipping of non-directories
- Introduces a DirectoryListMode, similar to SpanMode
- Uses low-overhead channels to reduce the communication overhead between the calling thread and the worker thread that calls the OS
- Adds FileInfo.path to properly support the new recursive directory iteration schemes

Note that this should only be merged once #241 is merged and this is rebased.

### DIFF
--- a/dub.sdl
+++ b/dub.sdl
@@ -13,9 +13,11 @@ mainSourceFile "source/vibe/appmain.d"
 
 configuration "winapi" {
 	subConfiguration "eventcore" "winapi"
+	versions "Windows7"
 }
 configuration "winapi-optlink" {
 	subConfiguration "eventcore" "winapi-optlink"
+	versions "Windows7"
 }
 configuration "epoll" {
 	subConfiguration "eventcore" "epoll"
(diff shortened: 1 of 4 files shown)


l-kramer (approved):

Geod24:
This broke Vibe.d on Musl. First, because `AT_SYMLINK_NOFOLLOW` was not defined in druntime ([bug that's on me](https://github.com/dlang/druntime/pull/3333)), and second, because `st_mtimensec` and `st_ctimensec` are not POSIX, they are extensions.

s-ludwig:
I just had a look and it would really be nice to unify the Druntime declarations across architectures as far as possible (although that is obviously not an immediate fix). Most systems are actually binary compatible for the time fields, but definitions range from `timespec` over just `time_t` to `time_t`+nsecs with either underscores prefixed or not.

s-ludwig:
Fix: #249

Geod24:
> Do you want to go ahead and add a musl based CI job? Granted, not much is happening at the OS interface these days, but since I don't personally use it, that's the only sane way to avoid it breaking.

I tried that a while ago with Vibe.d and dub, but had some issues, can't remember why. Will take another look.

Geod24:
> I just had a look and it would really be nice to unify the Druntime declarations across architectures as far as possible (although that is obviously not an immediate fix)

It would be nice if we could just parse the system headers.

PR #249, opened 2021-01-18
Fix compilation of the directory iteration code on additional POSIX based platforms

s-ludwig:
See #242

### DIFF
--- a/source/vibe/core/file.d
+++ b/source/vibe/core/file.d
@@ -1122,6 +1122,11 @@ version (Posix) {
 		static if (!is(typeof(AT_SYMLINK_NOFOLLOW)))
 			enum AT_SYMLINK_NOFOLLOW = 0x0020;
 	}
+
+	version (CRuntime_Musl) {
+		static if (!is(typeof(AT_SYMLINK_NOFOLLOW)))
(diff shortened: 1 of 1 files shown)


Geod24:
```
source/vibe/core/file.d(1065,46): Error: undefined identifier `AT_SYMLINK_NOFOLLOW`
source/vibe/core/file.d(1048,17): Error: undefined identifier `timespec`
source/vibe/core/file.d(1059,12): Error: static assert:  "Found no nanoseconds fields in struct stat"
source/vibe/core/file.d(1072,24):        instantiated from here: `getTimeField!"st_mtim"`
dmd failed with exit code 1.
```

https://github.com/vibe-d/vibe-core/pull/250

I fixed the `AT_SYMLINK_NOFOLLOW` upstream but you'll need a workaround for the time being.

s-ludwig:
Cherry picked the workflow commit from #250 and fixed another error.

s-ludwig:
Okay, not sure what was going on with `timespec`, but it's now down to this:
```
Linking...
cc: No such file or directory
Error: linker exited with status 255
```

s-ludwig:
Okay, let me just drop the workflow commit, so that this can get merged and then #250 is the one that adds the workflow...

PR #253, opened 2021-02-24
Remove a fully-qualified expression that relies on a DMD bug

BorisCarvajal:
Qualified name `vibe.core.sync.Monitor` is not really available just `Monitor`.
However, for some reason `Monitor` is conflicting with `Object.Monitor` so I had to rename it.

* [DMD bug](https://github.com/dlang/dmd/pull/12215)

### DIFF
--- a/source/vibe/core/taskpool.d
+++ b/source/vibe/core/taskpool.d
@@ -10,7 +10,7 @@ module vibe.core.taskpool;
 import vibe.core.concurrency : isWeaklyIsolated;
 import vibe.core.core : exitEventLoop, logicalProcessorCount, runEventLoop, runTask, runTask_internal;
 import vibe.core.log;
-import vibe.core.sync : ManualEvent, Monitor, createSharedManualEvent, createMonitor;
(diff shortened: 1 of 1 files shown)


source/vibe/core/taskpool.d
```diff
@@ -10,7 +10,7 @@ module vibe.core.taskpool;
 import vibe.core.concurrency : isWeaklyIsolated;
 import vibe.core.core : exitEventLoop, logicalProcessorCount, runEventLoop, runTask, runTask_internal;
 import vibe.core.log;
-import vibe.core.sync : ManualEvent, Monitor, createSharedManualEvent, createMonitor;
+import vibe.core.sync : ManualEvent, CoreSyncMonitor = Monitor, createSharedManualEvent, createMonitor;
```
PetarKirov, line 13:
```suggestion
import vibe.core.sync : ManualEvent, VibeSyncMonitor = Monitor, createSharedManualEvent, createMonitor;
```

Otherwise it looks like `vibe.core.sync` is publicly importing something from druntime.

Issue #86, opened 2018-07-24
Resource leak in createTimer


SirNickolas:
```d
int main() nothrow @safe {
    import core.time;
    import vibe.core.core;

    runTask({
        while (true) {
            auto timer = setTimer(2.seconds, { assert(false); }, true);
            sleep(1.msecs);
            timer.stop();
        }
    });
    return runEventLoop();
}
```

This keeps consuming memory. After a minute, the program crashes with SIGABRT. Having glanced at the [source](https://github.com/vibe-d/vibe-core/blob/fae7d3e93d00d9636632aa0acf9ebc19ed9f4a34/source/vibe/core/core.d#L783), I suspect the leak to be caused by a task continuing to wait for a (stopped) timer indefinitely.

wilzbach:
Isn't everything that happens after an AssertError undefined behavior?
Can you also reproduce this without the `assert(0)`?

SirNickolas:
I put that `assert` to only notice that the callback isn't executed. Timer's period is `2.seconds`, and it is cancelled after `1.msecs`.

wilzbach:
Sorry only glanced briefly on it from my phone. Please ignore my comment then.

PR #255, opened 2021-03-10
Improve documentation for setTimer and sleep

Geod24:
Fairly simple, but I had to test it to figure out the behavior, instead of just looking up the docs.

### DIFF
--- a/source/vibe/core/core.d
+++ b/source/vibe/core/core.d
@@ -851,6 +851,22 @@ void switchToTask(Task t, TaskSwitchPriority priority)
 	wait time, in contrast to $(D core.thread.Thread.sleep), which shouldn't be
 	used in vibe.d applications.
 
+	Repeated_sleep:
+	  As this method creates a new `Timer` every time, it is not recommended to
+	  use it in a tight loop. For functions that calls `sleep` frequently,
+	  it is preferable to instantiate a single `Timer` and reuse it,
+	  as shown in the following example:
+	  ---
+	  void myPollingFunction () {
+		  Timer waiter = createTimer(null); // Create a re-usable timer
+		  while (true) {
+			  // Your awesome code goes here
+			  timer.rearm(timeout, false);
+			  timer.wait();
+		  }
+	  }
+	  ---
+
 	Throws: May throw an `InterruptException` if the task gets interrupted using
 		`Task.interrupt()`.
 */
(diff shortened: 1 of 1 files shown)


source/vibe/core/core.d
```diff
@@ -877,18 +892,64 @@ unittest {
...
+    int main (string[] args) {
+        Config conf = parseArguments(&args);
+        setTimer(0.msecs, () => setup(conf));
+        return runApplication(&args);
+    }
+    ---
+
+    As timers only run when the event loop is running, and `runApplication`
+    will start the event loop, `setupTimer` will just register the closure
+    that calls `setup` to be executed as soon as the event loop is ready.
```
s-ludwig, line 922: This is not the full truth, as any of vibe.d's "blocking" functions will run a temporary event loop. So as long as the thread is not stuck with a pure CPU load or `Thread.sleep`, the timer should be invoked at the right time regardless of `runApplication`.

source/vibe/core/core.d
```diff
@@ -877,18 +892,64 @@ unittest {
...
+    come at a surprise in code that doesn't handle request.
+    If this is a problem, the solution is usually to either explicitly give
+    control to the event loop (by calling `yield`) or ensuring operations are
+    asynchronous (e.g. call functions from `vibe.core.file` instead of `std.file`).
 
 	Params:
 		timeout = Determines the minimum amount of time that elapses before the timer fires.
-		callback = If non-`null`, this delegate will be called when the timer fires
+		callback = A delegate to be called when the timer fires. Can be `null`,
+                   in which case the timer will not do anything.
```
s-ludwig, line 948: Might make sense to mention the possibility to use `wait()` on a timer with no callback set here.

source/vibe/core/core.d
```diff
@@ -877,18 +892,64 @@ unittest {
 
 
 /**
-	Returns a new armed timer.
-
-	Note that timers can only work if an event loop is running, explicitly or
-	implicitly by running a blocking operation, such as `sleep` or `File.read`.
+	Creates a new timer, that will fire `callback` after `timeout`
+
+    There are two kinds of timer: one-off timers and periodic timers.
```
s-ludwig, line 897: Strictly speaking, the same kind of timer can be dynamically configured for those to two modes (i.e. a timer created as one-shot can later be rearmed as periodic and vice-versa).

source/vibe/core/core.d
```diff
@@ -851,6 +851,21 @@ void switchToTask(Task t, TaskSwitchPriority priority)
 	wait time, in contrast to $(D core.thread.Thread.sleep), which shouldn't be
 	used in vibe.d applications.
 
+    Repeated_sleep:
+      As this method creates a new `Timer` every time, it is not recommended to
+      use it in a loop. For this purpose, prefer to instantiate a single `Timer`
```
s-ludwig, line 856: I'd slightly reword this to something like "tight loop" or "called very frequently", as the creation overhead will usually only matter once the timeout enters millisecond territory. Might make sense to change this to maintain a small stack of reusable sleep timers per thread to avoid this issue.

source/vibe/core/core.d
```diff
@@ -877,18 +892,64 @@ unittest {
...
+	Creates a new timer, that will fire `callback` after `timeout`
+
+    There are two kinds of timer: one-off timers and periodic timers.
+    One-off timers fire only once, after a specific amount of time,
+    while periodic timer fire at a regular interval.
+
+    One-off_timers:
+    One-off timers can be used for performing a task after a specific delay,
+    or for the initial setup of an application. For example, the following
+    is a common pattern for a Vibe.d application:
```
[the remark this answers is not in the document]
Geod24: Hum, right. Actually one primitive I am missing from `std.concurrency` is `schedule`, that is, a `runTask` that doesn't switch immediately, which we currently use `Timer`  with a 0 timeout for.

An example use case: We get data, that we need to propagate. Once we validated it, we schedule a network request to our peer, then do other operations (updating local state, storing on disk, etc...).
When we used `runTask` originally, what would happen was that the nodes would DOS themselves, because they would immediately propagate it, then other nodes would receive it and immediately propagate it, and we might get the new propagation back before we're even done processing the data we originally received (once the processing is done, we ignore duplicated data).

I'll update the example with the `schedule` approach in mind.

source/vibe/core/core.d
Geod24: Indeed, it is more of a semantic classification. I'll try to rephrase.

Geod24:
> BTW, not sure if this is related to your experiences w.r.t. the preemption topic, but I still haven't found out why timers often tend to sleep considerably longer than they should (up to 200ms) when run inside of a VM, especially on a Mac. It seems to be related to the timeout handling of the OS event API itself, but so far I didn't attempt to profile this to get a more precise insight.

Accuracy of timers don't matter *so* much for us ATM. We could take a 5s jitter and not notice it.

For reference, I was trying to debug which of the dozen of periodic timers we use was not properly cleaned up and leading to the "Leaking event loop driver [...]" message. While instrumenting `vibe-core` with some stack printings, I was flooded by messages, because someone was doing a `sleep(10.msecs)` in a `while (1)`. Refactoring it into a function that was called by a periodic timer yielded questions as to whether it was safe to do so, given concerns of re-entrancy.

Which brings me to my next suggestion: It would be great to have a version / mode to record the stack trace of timer (or other user-created events) creations, and display it in addition to the "leaking" message. I'm working towards making it easy to do in user code by exposing an API for it in druntime (https://github.com/dlang/druntime/tree/master/src/core/internal/backtrace) but I need to finalize it.

Geod24:
Comments addressed.

source/vibe/core/core.d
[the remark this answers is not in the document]
s-ludwig: True, I had the need for that in the past, too, but always resorted to `yield()` at the start of the task, because it seemed to be a niche case. But I introduced `TaskSettings` a while ago, that would lend itself very well for a `deferred` flag.

s-ludwig:
> Refactoring it into a function that was called by a periodic timer yielded questions as to whether it was safe to do so, given concerns of re-entrancy.

Speaking of which. looking at the code, I think this part got broken by the fix for #86 and the consecutive fixup in 03ffe70dc1f583276b729bd7b5baa68bbdca0906. The current implementation will happily run multiple callbacks in parallel, which was definitely not intended from my side and I'll look into a fix.

Geod24:
I'll wait for the fix to update the documentation then :D

s-ludwig:
Opened #256

Geod24:
Updated with a re-entrancy section.

s-ludwig (approved):

PR #257, opened 2021-03-14
Remove noisy trace log

Geod24:
```
When a user enable trace logs, the 'emit shared done' and 'lw ...' messages
are displayed often, even when nothing user-related is happening.
This makes the log a little less verbose so that one can focus
on more user-relevant informations, such as the content of HTTP headers.
```

@s-ludwig : One thing I'm really missing in Vibe.d's logging capabilities is categorization. For example, take a look at Ocean's logger:
https://github.com/sociomantic-tsunami/ocean/blob/b230b02c63d946d686c344fda63cd4ef4a003c2c/src/ocean/util/log/Logger.d#L12-L29

The output for our app [looks like this](https://github.com/bpfkorea/agora/blob/d5d2928f27fc07204213eeef836b404ed6de7379/source/agora/consensus/protocol/Nominator.d#L417):
```
2021-03-14 10:40:36,495 Info [agora.consensus.protocol.Nominator] - agora.consensus.protocol.Nominator.Nominator.nominate(): Proposing tx set for slot 240
```

Would be good to have something similar in Vibe.d.

### DIFF
--- a/source/vibe/core/sync.d
+++ b/source/vibe/core/sync.d
@@ -1025,15 +1025,12 @@ struct ManualEvent {
 	shared nothrow @trusted {
 		import core.atomic : atomicOp, cas;
 
-		() @trusted { logTrace("emit shared %s", cast(void*)&this); } ();
-
 		auto ec = atomicOp!"+="(m_emitCount, 1);
 		auto thisthr = Thread.getThis();
 
 		ThreadWaiter lw;
 		auto drv = eventDriver;
 		m_waiters.lock.active.filter((ThreadWaiter w) {
-			() @trusted { logTrace("waiter %s", cast(void*)w); } ();
 			if (w.m_driver is drv) {
 				lw = w;
 				lw.addRef();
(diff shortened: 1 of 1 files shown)


s-ludwig:
I'd suggest to just put them inside `debug (VibeMutexLog)` for now, as they are sometimes useful do debug concurrency issues. BTW, those HTTP request log calls mentioned should actually be `logDebugV` instead of `logTrace`, which is meant to be for messages that indicate the control flow of the program - I should do a pass through the sources at some point to clean that up (especially since I just recently looked at those in a log output myself).

Geod24:
Having `version` statement is a step forward, but it doesn't quite help debug real-world apps, because it requires compilation to enable logs. In addition to the hierarchy of logger mentioned above, we had (at Sociomantic) the ability to [configure them selectively](https://github.com/sociomantic-tsunami/ocean/blob/8c0b7bdb15f6b727a73e45c31c1f48cb11ac5c21/src/ocean/util/log/Config.d#L1-L69) via config file, which allowed us to do things without recompilation. Some apps had real-time requirements and couldn't be restarted, so we also introduced a control interface via unix socket, allowing to dynamically change the log level used.

s-ludwig:
Sure, that makes sense, the `debug` would just be the right immediate solution.

So we have the calling module name available in `doLog()`, which would make it simple to add a trie-like structure for matching disabled or enabled modules. Using a simplified version (to support just fixed strings or terminating wild cards) of the `MatchTree` used by `HTTPRouter` for that sounds straight forward. Using explicitly named scopes like Ocean does would of course also be a possibility, although I really like the simplicity at the call site of the module based solution.

PR #273, opened 2021-05-13
Remove deprecated aliases and functions in GenericPath

Geod24:
Those deprecations were added in commit 07e077a0098be467496f43199de178a4a0dcb6b9,
dated 2017-06-19 and part of the very first release (v1.0.0, 2017-07-10).
The toString deprecation is the most prominent one, as it triggers under conditions
which the user can't fix, such as when calling format or similar.

CC @s-ludwig , our build logs are pristine save for this:
Screen Shot 2021-05-13 at 11 15 28

Because we use `-checkaction=context` which in turns calls `toString`.

### DIFF
--- a/source/vibe/core/path.d
+++ b/source/vibe/core/path.d
@@ -203,11 +203,6 @@ alias InetPath = GenericPath!InetPathFormat;
 version (Windows) alias NativePath = WindowsPath;
 else alias NativePath = PosixPath;
 
-deprecated("Use NativePath or one the specific path types instead.")
-alias Path = NativePath;
-deprecated("Use NativePath.Segment or one the specific path types instead.")
-alias PathEntry = Path.Segment;
-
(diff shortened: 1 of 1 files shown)


s-ludwig:
This unfortunately has to be postponed to 2.0.0, as the policy is to allow deprecations within a major release, but removing APIs is considered a breaking change.

We should probably do something about the `toString` situation, though, although its semantics are debatable. It could be argued that it should return the encoded form ("foo%20bar") instead of the human readable one ("foo bar"), and this was also the reason why I wanted to remove it - it's just ambiguous in a dangerous way.

Geod24:
> We should probably do something about the toString situation, though

Any suggestion ?

s-ludwig:
@Geod24 Can you do a quick rebase here? I'd start to integrate 2.x changes on master now and branch off a 1.x branch, if necessary.

Geod24:
Rebased

PR #279, opened 2021-05-19
Deprecate throwing tasks

s-ludwig:
Not catching an exception thrown from a task function results in the task getting silently terminated. This frequently results in hidden bugs where certain background processes just hang.

Instead of trying to do something more elaborate, such as rethrowing exceptions from `join` (which is not safely possible for worker tasks and cannot be statically enforced anyway), the plan is to simply disallow throwing task functions. Anything else can be layered on top if required.

### DIFF
--- a/source/vibe/core/channel.d
+++ b/source/vibe/core/channel.d
@@ -371,10 +371,12 @@ unittest {
 	void test(ChannelPriority prio)
 	{
 		auto ch = createChannel!int(ChannelConfig(prio));
-		runTask({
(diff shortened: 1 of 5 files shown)


l-kramer (approved):

Geod24:
@s-ludwig : Would it make sense to also deprecate `setTimer` that accepts a non-`nothrow` callback ?
Currently the exception is caught and logged, but it feels like the same reasoning that applies to `runTask` should apply to `setTimer`.

PR #285, opened 2021-06-10
print interface conformance errors with FQN

WebFreak001:
makes it obvious in which modules types are for quicker troubleshooting.

The `from T` isn't as accurate as it could be (doesn't show inheritance of interface / only shows the passed interface and not anything it was inherited from) but it's good enough to find issues.


Geod24:
Needs a rebase (and a fix).

WebFreak001:
needs a rebase? looks up-to-date here - working on fix

PR #292, opened 2021-08-24
Fix possibly hanging process after SIGINT/SIGTERM.

s-ludwig:
Ensures that the thread local event loop gets terminated, even if it is currently in an infinite wait state.

Fixed vibe-d/vibe.d#2596

### DIFF
--- a/source/vibe/core/core.d
+++ b/source/vibe/core/core.d
@@ -2058,6 +2058,13 @@ nothrow {
 	catch (Throwable th) {
 		logDebug("Failed to notify for event loop exit: %s", th.msg);
 	}
+
+	// Stop the event loop of the current thread directly instead of relying on
+	// the st_term mechanic, as the event loop might be in a waiting state, so
+	// that no tasks can get scheduled before an event arrives. Explicitly
+	// exiting the event loop on the other hand always terminates the wait state.
+	if (auto drv = tryGetEventDriver())
+		drv.core.exit();
 }
 
 private extern(C) void onBrokenPipe(int signal)


Geod24:
Tested, confirmed fixed, thanks!

Geod24:
@s-ludwig : Not wanting to move the goal post, but since we're on the topic of shutdown:

Currently, it seems that Vibe do not shutdown listeners itself. Most of our examples use things like `listenHTTP`, e.g. [the app_skeleton](https://github.com/vibe-d/vibe.d/blob/5b91bb13909580aa115d64a449bdebd355c4de00/examples/app_skeleton/source/app.d#L27), but we should actually be recommending `auto listener = listenHTTP(...);` so that the user can shut down the connection.

However, even that is not enough. If one really want to get rid of the shutdown messages, one has to disable the default signal handlers and provide their own implementation. For some, it is necessary anyway, as they might need to perform some non-Vibe-related cleanup on shutdown. See [our example](https://github.com/bosagora/agora/blob/d0de2335ea2fc6004b1fdceabf0975769b3c224f/source/agora/node/main.d#L214-L264), where we need to call the node's `shutdown()` procedure.

It's not a blocking issue per se (it's possible to do what we want with the current API), but I feel the API fails to address a common, and IMO basic, need. There's a few ways this could be addressed, but I was wondering first if you were aware of this / had some thoughts on it ?

s-ludwig:
I had this topic on my mind for some time and I definitely agree with the `auto listener = listenHTTP((...)`. But since we don't do the `shared static this` approach anymore, I think that we don't really have to change the shutdown procedure itself that much\* or add new APIs. It should be sufficient to simply move any shutdown procedures to after the `runApplication`/`runEventLoop` call.

\* What comes to mind is that currently the event loops of all threads are exited, whereas only the main one should be. The main thread would then have to make sure that other threads are shut down properly.

Geod24:
Hum, I'll give it a shot in our app, to see if that currently works.

> But since we don't do the `shared static this` approach anymore [...]

Well then if I ever get free time I'll definitely look at https://github.com/vibe-d/vibe.d/issues/2370 then :)

PR #295, opened 2021-09-13
Minor compatibility fixes for upcoming release

Geod24:
I was going through the deprecated symbol and see the v2.069 comment, hence the first commit.
The second commit was found because we use `-checkaction=context`, which triggers the following error:
```
/Users/runner/hostedtoolcache/dc/ldc2-36198b36/x64/ldc2-36198b36-osx-x86_64/bin/../import/core/internal/dassert.d(194,36): Error: cannot implicitly convert expression `v.opCast()` of type `bool` to `const(ManualEvent)`
/Users/runner/hostedtoolcache/dc/ldc2-36198b36/x64/ldc2-36198b36-osx-x86_64/bin/../import/core/internal/dassert.d(449,30): Error: template instance `core.internal.dassert.miniFormat!(shared(ManualEvent))` error instantiating
/Users/runner/hostedtoolcache/dc/ldc2-36198b36/x64/ldc2-36198b36-osx-x86_64/bin/../import/core/internal/dassert.d(334,85):        instantiated from here: `formatMembers!(TaskFiber)`
/Users/runner/hostedtoolcache/dc/ldc2-36198b36/x64/ldc2-36198b36-osx-x86_64/bin/../import/core/internal/dassert.d(527,19):        3 recursive instantiations from here: `miniFormat!(const(TaskFiber)[])`
/Users/runner/hostedtoolcache/dc/ldc2-36198b36/x64/ldc2-36198b36-osx-x86_64/bin/../import/core/internal/dassert.d(75,49):        instantiated from here: `miniFormatFakeAttributes!(const(TaskFiber)[])`
submodules/vibe-core/source/vibe/internal/array.d(389,10):        instantiated from here: `_d_assert_fail!(TaskFiber[])`
submodules/vibe-core/source/vibe/core/core.d(1832,2):        instantiated from here: `FixedRingBuffer!(TaskFiber, 0LU, true)`
```

### DIFF
--- a/source/vibe/core/log.d
+++ b/source/vibe/core/log.d
@@ -408,8 +408,7 @@ final class FileLogger : Logger {
 			case Format.threadTime:
 				dst.put('[');
 				auto tm = msg.time;
-				static if (is(typeof(tm.fracSecs))) auto msecs = tm.fracSecs.total!"msecs"; // 2.069 has deprecated "fracSec"
-				else auto msecs = tm.fracSec.msecs;
+			    auto msecs = tm.fracSecs.total!"msecs";
 				m_curFile.writef("%d-%02d-%02d %02d:%02d:%02d.%03d ", tm.year, tm.month, tm.day, tm.hour, tm.minute, tm.second, msecs);
 
 				if (msg.threadName.length) dst.put(msg.threadName);
--- a/source/vibe/core/sync.d
+++ b/source/vibe/core/sync.d
@@ -840,7 +840,7 @@ struct LocalManualEvent {
 		}
 	}
 
-	bool opCast() const nothrow { return m_waiter !is null; }
+	bool opCast (T : bool) () const nothrow { return m_waiter !is null; }
 
 	/// A counter that is increased with every emit() call
 	int emitCount() const nothrow { return m_waiter.m_emitCount; }
@@ -1041,7 +1041,7 @@ struct ManualEvent {
 	}
 
 	deprecated("ManualEvent is always non-null!")
-	bool opCast() const shared nothrow { return true; }
+	bool opCast (T : bool) () const shared nothrow { return true; }
 
 	/// A counter that is increased with every emit() call
 	int emitCount() const shared nothrow @trusted { return atomicLoad(m_emitCount); }


hewison-chris (approved):
LGTM

Geod24:
Self merging as it's trivial (and unbreaks our build).

Geod24:
That didn't actually work, because I forgot to pass `-checkaction=context` while testing.
Tracked it down and reported it as https://issues.dlang.org/show_bug.cgi?id=22300 .

Geod24:
@s-ludwig : Any plans to clean up the deprecated symbols ? This one has been in since v1.0.0, so it's been more than 4 years.

s-ludwig:
> Any plans to clean up the deprecated symbols ? This one has been in since v1.0.0, so it's been more than 4 years.

Actually removing the deprecated symbols needs to wait until 2.0.0 comes out. I don't have a plan/reason to go for 2.0.0 right now, though. If there is anything deprecated that *really* interferes with development, we could do that, but otherwise that would be quite unnecessary breakage.

Geod24:
> Actually removing the deprecated symbols needs to wait until 2.0.0 comes out.

Agreed, which is why I went with this approach. Unfortunately it didn't work out.

>  I don't have a plan/reason to go for 2.0.0 right now, though.

I think that major releases should be put out at a regular interval rather than on a need basis. Obviously Vibe.d & co are seeing uneven development activity so that's pretty hard to do, but I don't think that putting out a new major now (e.g. after 4 years) would be bad. I don't think we should have any FOMO when it comes to release (if we have regular releases, then a major change that doesn't make the cut will make the next one, not a big deal).

> If there is anything deprecated that really interferes with development, we could do that, but otherwise that would be quite unnecessary breakage.

We will probably just fork the repository, and have a patch that removes those deprecated symbols. We're in quite a niche development-wise (we use a few preview, checkaction=context, etc..), and since I reported the regression, it's probably going to be fixed upstream, although maybe not in time for the next release.

s-ludwig:
We could also use a `version` to explicitly remove the deprecated symbols here, to avoid the need for a fork.

Geod24:
Should've thought of this... Well I went with the fork approach so far: https://github.com/bosagora/vibe-core/commit/1aa6c05097bfa5cf667d53190da6aa2a19340c1f
It also solves the issue we have with deprecated symbols (https://github.com/vibe-d/vibe-core/pull/273), so it's a win, and since we're using `dub` only for building, not dependency resolution, it's trivial to do.

As mentioned above, it's reported as a regression, so it's eventually going to get fixed. Regarding the deprecations (`toString` and what you mentioned in the Vibe.d PR about `enum` members triggering a flood of deprecations on `to`), I think the DLF should establish a policy for this (e.g. do we ignore it, or work around it), I'll try to bring it up at the next meeting.

PR #299, opened 2021-10-15
Omit unreachable statement in `asynAwaitAny` for noreturn callbacks

MoonlightSentinel:
The trailing statements become unreachable when the callbacks return
`noreturn`, e.g. because they are known to always throw an exception.

Hence check the return type of the callables and only emit the
statements when the code is actually reachable.

---

Draft PR because this currently triggers an assertion failure in the respective backends of dmd and ldc


s-ludwig:
My feeling here is that we should (rather?) enforce `nothrow` for all callbacks, as skipping the `switchToTask` call will let the calling task starve. So the only real application that I'd see is implementing a test that contains an `assert(false)` in the respective callback body. Was that the intended use for this?

(BTW, note that the passed Windows tests were bogus, see #330)

PR #303, opened 2021-11-29
ConnectionPool and TaskMutex enhancements

omerfirmak:


Geod24:
First commit must go

omerfirmak:
> First commit must go

done

source/vibe/core/connectionpool.d
```diff
@@ -125,6 +125,42 @@ final class ConnectionPool(Connection)
...
+
+		Params:
+			conn = connection to remove from the pool
+	*/
+	void remove(Connection conn) @safe
+	{
+		foreach (idx, c; m_connections)
+			if (c is conn)
+			{
+				m_connections = m_connections[0 .. idx] ~ m_connections[idx + 1 .. $];
```
s-ludwig, line 139: Should probably also remove from `m_lockConn` to allow calling `add` later on. However, that would mean that the lock count must be asserted to be zero here, or `LockedConnection` would re-establish the entry with a wrong value.

source/vibe/core/connectionpool.d
```diff
@@ -125,6 +125,42 @@ final class ConnectionPool(Connection)
 		foreach (c; removed_conns)
 			disconnect_callback(c);
 	}
+
+	/** Removes an existing connection from the pool
```
s-ludwig, line 129: I think this needs to document the gotchas associated with using explicit `add`/`remove` calls, mainly telling when `remove` is allowed to be called (lock count zero or non-zero?) and potentially what this means when `add` is called with such a connection later on.

source/vibe/core/connectionpool.d
```diff
@@ -125,6 +125,42 @@ final class ConnectionPool(Connection)
...
+
+		Params:
+			conn = new connection to add to the pool
+
+		Returns:
+			success/failure
+	*/
+	bool add(Connection conn) @safe nothrow
+	{
+		assert((conn in m_lockCount) is null);
```
s-ludwig, line 155:
```suggestion
		assert((conn in m_lockCount) is null, "Attempt to add connection to pool that is already present");
```

s-ludwig (changes requested):

source/vibe/core/connectionpool.d
omerfirmak: > lock count must be asserted to be zero here

But it is most likely locked, since someone already has a reference to the connection and is trying to remove it from the pool.

Check the updated code.

source/vibe/core/connectionpool.d
omerfirmak: Done

source/vibe/core/connectionpool.d
s-ludwig: Okay, that should work!

s-ludwig (approved):

s-ludwig:
For some reason the tests didn't run, can you do a quick rebase to re-trigger?

Issue #282, opened 2021-05-29
vibe.core.process.pipe cannot call @system function core.stdc.stdio._handleToFD!()._handleToFD


SingingBush:
On Windows build with dmd 2.094 to 2.096 I'm seeing these errors with vibe-core 1.18.0:

```
vibe-core 1.18.0: building configuration "winapi"...
C:\Users\runneradmin\AppData\Local\dub\packages\vibe-core-1.18.0\vibe-core\source\vibe\core\process.d(608,28): Error: `@safe` function `vibe.core.process.pipe` cannot call `@system` function `core.stdc.stdio._handleToFD!()._handleToFD`
C:\hostedtoolcache\windows\dc\dmd-2.096.1\x64\dmd2\windows\bin\..\..\src\druntime\import\core\stdc\stdio.d(1823,9):        `core.stdc.stdio._handleToFD!()._handleToFD` is declared here
C:\Users\runneradmin\AppData\Local\dub\packages\vibe-core-1.18.0\vibe-core\source\vibe\core\process.d(609,29): Error: `@safe` function `vibe.core.process.pipe` cannot call `@system` function `core.stdc.stdio._handleToFD!()._handleToFD`
C:\hostedtoolcache\windows\dc\dmd-2.096.1\x64\dmd2\windows\bin\..\..\src\druntime\import\core\stdc\stdio.d(1823,9):        `core.stdc.stdio._handleToFD!()._handleToFD` is declared here
C:\hostedtoolcache\windows\dc\dmd-2.096.1\x64\dmd2\windows\bin\dmd.exe failed with exit code 1.
Error: Process completed with exit code 1.
```

SingingBush:
worth noting that this is with all dmd versions since 2.094 and also affects vibe-core 1.22.0:

```
C:\Users\runneradmin\AppData\Local\dub\packages\vibe-core-1.22.0\vibe-core\source\vibe\core\process.d(608,28): Error: `@safe` function `vibe.core.process.pipe` cannot call `@system` function `core.stdc.stdio._handleToFD!()._handleToFD`
C:\hostedtoolcache\windows\dc\dmd-2.098.1\x64\dmd2\windows\bin64\..\..\src\druntime\import\core\stdc\stdio.d(1876,9):        `core.stdc.stdio._handleToFD!()._handleToFD` is declared here
[...]

PR #305, opened 2022-01-13
additional Windows builds

SingingBush:
added more dmd versions for testing on Windows to highlight problem in #282


SingingBush:
closing as it would be better to merge #306

PR #311, opened 2022-03-22
Make parallelMap clean up resources deterministically.

s-ludwig:
Avoids relying on the GC to clean up inputs or outputs.

### DIFF
--- a/source/vibe/core/parallelism.d
+++ b/source/vibe/core/parallelism.d
@@ -144,7 +144,7 @@ auto parallelMap(alias fun, R)(R items, shared(TaskPool) task_pool, ChannelConfi
 	import std.algorithm : canFind, countUntil, move, remove;
 	import std.container.array : Array;
 	import std.range : enumerate;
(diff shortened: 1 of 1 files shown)


source/vibe/core/parallelism.d
```diff
@@ -200,6 +201,29 @@ auto parallelMap(alias fun, R)(R items, shared(TaskPool) task_pool, ChannelConfi
 
 	static struct Result {
 		private State* state;
```
PetarKirov, line 203: Any particular disadvantages to using `std.typecons.RefCounted` that you have in mind?

s-ludwig: Could work here, I just generally stay away from it due to all the shortcomings it used to have (not sure about the status quo).

s-ludwig: Changed to use `RefCounted` and appears to work fine.

l-kramer (approved):

PR #317, opened 2022-06-15
Fix deprecation of throwing `state.readex`

### LINKED ISSUE #316: Deprecation: scope variable `state` may not be thrown
https://github.com/vibe-d/vibe-core/blob/a646ede53bfabed69ebba9b5d91f774b27213c5f/source/vibe/core/stream.d#L168

CodeMyst:
My attempt at solving the deprecation message (#316). Not entirely sure if this is the right way to go about it, but it seems like it works and passes all the unit tests.

Closes #316

### DIFF
--- a/source/vibe/core/stream.d
+++ b/source/vibe/core/stream.d
@@ -87,7 +87,6 @@ ulong pipe(InputStream, OutputStream)(InputStream source, OutputStream sink,
 					size_t[bufcount] bufferFill;
 					// buffer index that is being read/written
 					size_t read_idx = 0, write_idx = 0;
-                    Exception readex;
 					bool done = false;
 					LocalManualEvent evt;
 					size_t bytesWritten;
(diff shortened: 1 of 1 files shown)


Geod24:
Switching to delegate means an extra allocation. I'm not sure that's what we want. Could you look into:
- Marking the parameter as `scope`;
- Passing a pointer to `Exception` to be assigned to ?

CodeMyst:
Passing the exception to the task works. Didn't pass a pointer to the Exception since it's already a class.

s-ludwig:
The problem here is that the assignment `readex = e` will only assign the exception reference to the parameter of the task callback, but not to the original stack variable. So this indeed needs to be changed to pass a pointer to `readex` to the task.

What I'm not sure about is why this would work with a free-standing variable, but not with a struct field. Seems to me like the propagation of the inferred `scope` is currently ill-defined.

PR #323, opened 2022-09-17
Fix iOS compilation.

l-kramer:


source/vibe/core/file.d
```diff
@@ -1229,7 +1229,12 @@ version (Posix) {
...
 	}
 
-	version (darwin) {
+	version (OSX) {
+		static if (!is(typeof(AT_SYMLINK_NOFOLLOW)))
+			enum AT_SYMLINK_NOFOLLOW = 0x0020;
+	}
+
+	version (iOS) {
 		static if (!is(typeof(AT_SYMLINK_NOFOLLOW)))
```
Geod24, line 1238: That looks redundant ?

s-ludwig: Do you mean the `static if` in particular, or the `darwin` -> `iOS`/`OSX`? Both look fine to me, considering that `darwin`<=>`OSX`.

Geod24: `darwin` should cover iOS and OSX, and the `AT_SYMLINK_NOFOLLOW` value is the same, so all this seem to do is excluding TvOS and other darwin-derived platform ?

s-ludwig: This has not been the case with my experiments, though. Dub treats its version of "darwin" like that, and of course that makes sense. The compiler though seems to treat is as identical to OSX - also, it is of course deprecated.

s-ludwig: Relevant code for LDC:
- https://github.com/ldc-developers/ldc/blob/39727641d55119ab3f364036ca1e275e880b8f35/driver/main.cpp#L813
- https://github.com/ldc-developers/ldc/blob/39727641d55119ab3f364036ca1e275e880b8f35/driver/main.cpp#L851-L854

Issue #324, opened 2022-10-04
Running Websocket server on a thread gets locked


grillo-delmal:
I've been having a conversation in another project where we are implementing a Websocket server inside a thread for a specific module that it can be optionally activated and deactivated during runtime to feed the application with data. The current implementation is as follows (removing the unnecessary details):

```d
class ThreadedReceiver {
private:
    ushort port = 8080;
    string bind = "0.0.0.0";

    bool isCloseRequested;
    Thread receivingThread;

public:
    ~this() {
        this.stop();
    }

    void handleConnection(scope WebSocket socket) {
        while (!isCloseRequested && socket.connected) {
	    // Receive data
        }
    }

    void receiveThread() {
        isCloseRequested = false;

        // ...

        HTTPServerSettings settings =  new HTTPServerSettings();
        settings.port = port;
        settings.bindAddresses = [bind];

        auto router = new URLRouter;
        router.get("/", handleWebSockets(&this.handleConnection));

        HTTPListener listener = listenHTTP(settings, router);

        while (!isCloseRequested) {
            runEventLoopOnce();
        }
        listener.stopListening();
    }
	
    void start() {
        // ...
        receivingThread = new Thread(&receiveThread);
        receivingThread.start();
    }

    void stop() {
        // ...
        isCloseRequested = true;
        receivingThread.join();
        receivingThread = null;
    }
    // ...

}
```

This implementation works as expected, but the problem is that when the stop function is called, since the [runEventLoopOnce](https://vibed.org/api/vibe.core.core/runEventLoopOnce) method waits indefinitely for data to process, if the socket is never connected then the app will lock on when the stop function is called, which is not ideal.

Reading the documentation, it seems that replacing the [runEventLoopOnce](https://vibed.org/api/vibe.core.core/runEventLoopOnce) method with the [processEvents](https://vibed.org/api/vibe.core.core/processEvents) could work too, but it overstresses the thread and the CPU in the process.

I think that this can be fixed if the timeout variable used in the scheduler is exposed to one of these methods, so I'll send a PR, but I'm interested to know if there is another way to handle this.

grillo-delmal:
Thank you for the advice, we ended up resolving the issue using TaskCondition.wait following the example in the documentation :)

I'll close this issue and leave the PR, you can discard it if you think that it's not needed.

PR #325, opened 2022-10-04
Expose timeout on runEventLoopOnce

grillo-delmal:
PR regarding issue #324

### DIFF
--- a/source/vibe/core/task.d
+++ b/source/vibe/core/task.d
@@ -834,11 +834,6 @@ package struct TaskScheduler {
 		This will execute scheduled tasks and process events from the
 		event queue, as long as possible without having to wait.
 
-		Params:
-			timeout = Maximum amount of time to wait for an event. A duration of
-				zero will cause the function to only process pending events. A
-				duration of `Duration.max`, if necessary, will wait indefinitely
-				until an event arrives.
 		Returns:
 			A reason is returned:
 			$(UL
@@ -852,7 +847,7 @@ package struct TaskScheduler {
 					but there were no pending events present.)
 			)
 	*/
-	ExitReason process(Duration timeout=0.seconds)
+	ExitReason process()
 	{
 		assert(TaskFiber.getThis().m_yieldLockCount == 0, "May not process events within an active yieldLock()!");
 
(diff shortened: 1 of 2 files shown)


s-ludwig:
I can see two potential issues with this extension:
- The loop in `TaskScheduler.process` can cause the call to continue indefinitely if any tasks are active. The code would have to be adjusted to either aim at an absolute timeout (initially record `MonoTime.currTime + timeout`) instead of continuously using the same value, or maybe even better to set the timeout to zero once it was hit for the first time.
- `processEvents` doesn't perform idle processing, in contrast to `runEventLoopOnce`, and thus usually needs to be paired with a call to `performIdleProcessing`. With a timeout present, this could lead to issues, depending on how exactly the timeout is implemented, as it might be impossible to properly interleave the `performIdleProcessing` calls.

The second point suggests to me that it probably makes more sense to introduce the `timeout` parameter to `runEventLoopOnce`  and `TaskScheduler.waitAndProcess` instead.

grillo-delmal:
Made the change you asked 👍 .

Honestly I went originally for the processEvents method because runEventLoopOnce has an assert blocking the timeout in the `waitAndProcess` method, so processEvents worked on my first try xD.

Just so you know, I already solved my issue with the advice you gave me, but this should work too.

s-ludwig (approved):

source/vibe/core/task.d
```diff
@@ -907,15 +913,17 @@ package struct TaskScheduler {
...
+		er = eventDriver.core.processEvents(timeout);
 		debug (VibeTaskLog) logTrace("Done: %s", er);
 		final switch (er) {
 			case ExitReason.exited: return ExitReason.exited;
 			case ExitReason.outOfWaiters:
 				if (!scheduledTaskCount)
 					return ExitReason.outOfWaiters;
 				break;
-			case ExitReason.timeout: assert(false, "Unexpected return code");
+			case ExitReason.timeout:
```
s-ludwig, line 924:
```suggestion
			case ExitReason.timeout:
```

There is a trailing white space here that makes the basic style check trip.

PR #329, opened 2023-01-16
Fix almost all remaining scope related warnings

s-ludwig:
Still has one warning that complains about passing a `scope` inferred argument to `runWorkerTask`. The cause seems to be the same as for #316, namely, all members of the passed struct value are also treated as if they had a `scope` in front of them. Since there is a class reference contained inside the struct, the memory of the class instance is suddenly also treated as `scope`.

**This PR contains breaking changes in `OutputStream` and `vibe.core.process`, which will require bumping to 2.0.0 for the next release**

### DIFF
--- a/source/vibe/core/file.d
+++ b/source/vibe/core/file.d
@@ -732,7 +732,7 @@ struct FileStream {
 		//       be relied upon. For this reason, we MUST use the uninterruptible
 		//       version of asyncAwait here!
 		auto res = asyncAwaitUninterruptible!(FileIOCallback,
-			cb => eventDriver.files.read(m_fd, ctx.ptr, dst, mode, cb)
+			cb => eventDriver.files.read(m_fd, ctx.ptr, () @trusted { return dst; } (), mode, cb)
 		);
 		ctx.ptr += res[2];
 		enforce(res[1] == IOStatus.ok, "Failed to read data from disk.");
--- a/source/vibe/core/file.d
+++ b/source/vibe/core/file.d
@@ -745,13 +745,13 @@ struct FileStream {
 		assert(ret == dst.length, "File.read returned less data than requested for IOMode.all.");
 	}
 
-	size_t write(in ubyte[] bytes, IOMode mode)
+	size_t write(scope const(ubyte)[] bytes, IOMode mode)
 	{
 		// NOTE: cancelWrite is currently not behaving as specified and cannot
 		//       be relied upon. For this reason, we MUST use the uninterruptible
 		//       version of asyncAwait here!
 		auto res = asyncAwaitUninterruptible!(FileIOCallback,
-			cb => eventDriver.files.write(m_fd, ctx.ptr, bytes, mode, cb)
+			cb => eventDriver.files.write(m_fd, ctx.ptr, () @trusted { return bytes; } (), mode, cb)
 		);
 		ctx.ptr += res[2];
 		if (ctx.ptr > ctx.size) ctx.size = ctx.ptr;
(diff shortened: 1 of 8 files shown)


s-ludwig:
Moved the test related changes to #330.

source/vibe/core/file.d
```diff
@@ -736,7 +736,7 @@ struct FileStream {
 		//       be relied upon. For this reason, we MUST use the uninterruptible
 		//       version of asyncAwait here!
 		auto res = asyncAwaitUninterruptible!(FileIOCallback,
-			cb => eventDriver.files.read(m_fd, ctx.ptr, dst, mode, cb)
+			cb => eventDriver.files.read(m_fd, ctx.ptr, () @trusted { return dst; } (), mode, cb)
```
Geod24, line 739: Why the trusted ? Isn't this just essentially bypassing every scope checks ?

source/vibe/core/log.d
```diff
@@ -423,9 +423,12 @@ final class FileLogger : Logger {
 
 	override void put(scope const(char)[] text)
 	{
-		static if (__VERSION__ <= 2066)
+		static if (__VERSION__ <= 2066) {
```
Geod24, line 426: I'd say we just drop this, we don't support this version anymore...

source/vibe/core/log.d
```diff
@@ -423,9 +423,12 @@ final class FileLogger : Logger {
...
 			() @trusted { m_curFile.write(text); } ();
-		else m_curFile.write(text);
+		} else {
+			// NOTE: (DMD 2.101.2) File.write parameters are not annotated scope
```
Geod24, line 429: Did you raise an issue ?

source/vibe/internal/async.d
```diff
@@ -138,15 +138,15 @@ void asyncAwaitAny(bool interruptible, Waitables...)(string func = __FUNCTION__)
 				};
 
 				debug(VibeAsyncLog) logDebugV("Starting operation %%s", %1$s);
-				alias WR%1$s = typeof(Waitables[%1$s].wait(callback_%1$s));
-				static if (is(WR%1$s == void)) Waitables[%1$s].wait(callback_%1$s);
-				else auto wr%1$s = Waitables[%1$s].wait(callback_%1$s);
+				alias WR%1$s = typeof(Waitables[%1$s].wait(() @trusted { return callback_%1$s; } ()));
+				static if (is(WR%1$s == void)) Waitables[%1$s].wait(() @trusted { return callback_%1$s; } ());
+				else auto wr%1$s = Waitables[%1$s].wait(() @trusted { return callback_%1$s; } ());
```
Geod24, line 143: Could this end up trusting user code ?

source/vibe/core/file.d
s-ludwig: There is no way around this, unfortunately. The `read` function invokes the callback asynchronously, so it definitely isn't `scope` - instead, `asyncAwaitUninterruptible` ensures that the callback gets invoked before it returns through its program logic. This looks like prime example where `@trusted` is needed.

source/vibe/core/log.d
s-ludwig: Yeah, I'll remove that.

source/vibe/core/log.d
s-ludwig: No

source/vibe/internal/async.d
s-ludwig: I don't think so, the callbacks here are all confined to vibe-core's own source code and any user code/callbacks are layered on top.

Geod24:
@s-ludwig : What timeline are you thinking about for the next release ?

PR #341, opened 2023-02-27
Deprecate GenericPath.Segment.

s-ludwig:
This prepares the code for a full removal of GenericPath.Segment and is supposed to be included in the next 1.x.x release. In the 2.x.x branch, the removal will be completed and Segment2 will be renamed to Segment (keeping an alias for some backwards compatibility).

Note that this is a cherry pick from master (a460e7952a9358ae4b7c7a7447fd7af1b17a5164).

### DIFF
--- a/.github/workflows/musl.yml
+++ b/.github/workflows/musl.yml
@@ -3,10 +3,10 @@ name: Musl tests
 on:
   pull_request:
     branches:
-      - master
+      - 1.x.x
   push:
     branches:
-      - master
+      - 1.x.x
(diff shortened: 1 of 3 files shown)


Geod24:
So we're asking people to rename from `Segment` to `Segment2` then back to `Segment`, or am I missing something ?

s-ludwig:
Eventually, yes. This is something for the more distant future, though. There was unfortunately no way to make this transition with forward *or* backward compatibility without going through a separate type.

PR #342, opened 2023-02-28
Remove GenericPath.SegmentOld and associated code.

s-ludwig:
This completes the migration to Segment2 by renaming it to Segment, while keeping a backwards compatibility alias. Any code that used Segment2 will continue to work as-is, while code based on Segment may need some adjustments to explicitly convert the .name property to a string.


Geod24 (approved):
Nice

s-ludwig:
I've tagged 2.0.0-beta.1 now - if that turns out without issues over the next few days, I think we can go 2.0.0 final.

PR #345, opened 2023-03-07
Deprecation fixes

s-ludwig:


source/vibe/core/stream.d
```diff
@@ -35,7 +35,7 @@ public import eventcore.driver : IOMode;
 		The actual number of bytes written is returned. If `nbytes` is given
 		and not equal to `ulong.max`, íts value will be returned.
 */
-ulong pipe(InputStream, OutputStream)(InputStream source, OutputStream sink,
+ulong pipe(InputStream, OutputStream)(scope InputStream source, scope OutputStream sink,
```
Geod24, line 38: This is supposed to be inferred but I wouldn't trust DMD's `scope` impl :shrug:

Geod24 (approved):

source/vibe/core/stream.d
s-ludwig: There are definitely some strange corner cases left. I stopped fixing warnings in vibe.d, because there are just too many that just seem wrong and pointless. I'll just have to find some time to file bug reports.

In this case, it could also be class references, though, so if I understand it right, the `scope` is still potentially doing something (if just pushing the warning/error down to where it is actually caused)?

PR #346, opened 2023-03-08
Fix GenericPath.Segment.toString()

s-ludwig:
The .to!string implementation fell back to typeof(R).stringof because the passed range was const. Using an explicit save() call fixes that.


l-kramer (approved):

Geod24:
Doesn't it just mean one `toString` is missing a `const` somewhere ?

s-ludwig:
It's the range interface that has a non-const `popFront`. The original range that is passed to `to!string` doesn't have a `toString()` method.

PR #347, opened 2023-03-15
Fixed typo so that testing on server is not actually testing on proxy

Tusanga:
"Test protocol implementation on server with and without forced disconnect" referenced the proxy address (l2.bindAddress) and not the server address (l1.bindAddress). This has been changed.


Geod24:
You can rebase.

PR #218, opened 2020-08-12
Fix resolveHost returning addresses that do not match the expected address family

Yoplitein:

PR #349, opened 2023-03-16
Ip transparent socket option

Tusanga:
Add support for IP_TRANSPARENT socket options. Only Linux provides this option. Related to eventcore pull #218.

### DIFF
--- a/source/vibe/core/net.d
+++ b/source/vibe/core/net.d
@@ -137,6 +137,10 @@ TCPListener listenTCP(ushort port, TCPConnectionDelegate connection_callback, st
 		sopts |= StreamListenOptions.reusePort;
 	else
 		sopts &= ~StreamListenOptions.reusePort;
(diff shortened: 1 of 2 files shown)


source/vibe/core/net.d
```diff
@@ -137,6 +137,12 @@ TCPListener listenTCP(ushort port, TCPConnectionDelegate connection_callback, st
 		sopts |= StreamListenOptions.reusePort;
 	else
 		sopts &= ~StreamListenOptions.reusePort;
+	version(linux) {
```
s-ludwig, line 140:
```suggestion
	static if (is(typeof(StreamListenOptions.ipTransparent))) {
```

That way we stay compatible with older eventcore versions and keep the platform knowledge a bit more confined to one place.

s-ludwig (approved):

PR #350, opened 2023-03-16
Speed up InterfaceProxy compilation by avoiding `format` calls

s-ludwig:


Geod24:
Speed up is quite noticeable.

s-ludwig:
LDC's `-ftime-trace` FTW! Although I can't use it for normal dev work right now due to ldc-developers/ldc#4354, the time trace feature in conjunction with the CTFE output in 1.32.0 is absolutely invaluable in tracking these down.

PR #355, opened 2023-04-18
Added boolean return value to lowerPrivileges

Tusanga:


Tusanga:
Is this a timeout for (macOS-latest, dmd-latest, x86_64) or is there an error in new code?

s-ludwig:
Not sure what happened there, it doesn't look like a timeout. But it also can't really be related to the changes here, so let's ignore it...

s-ludwig:
What is the use case you have in mind for the return value? I'm asking because it seems not really clear whether the return value for the `uname == "" && gname == ""` case should be `true` or `false`. Depending on the situation, this could be something that is not wanted, or something that works as requested.

When looking at the code as is stands, it seems like it might be a better idea to forbid that case altogether (e.g. throw an exception) and move the warning message to the `lowerPrivileges` overload that takes no arguments. There just doesn't seem to be a point in explicitly passing empty arguments, fully expecting nothing to happen.

Tusanga:
Background: I have experienced accidentally started a service as the wrong not-root user only to find out later (not at  startup but during later request processing) that I can't open specific files. I would like to be able to prevent that with a false return value immediately after the call to lowerPrivileges. With a check one can detect if the process runs as a non-root. If uname == "" && gname == "" I expect true as I do not want act on that. Always starting the service as root allows me during debugging to bind to ports below and above 1024 without having to change any code.

PR #359, opened 2023-09-19
TcpConnect.read(buffer,IOMode.once) should throw ReadTimeoutException

dushibaiyu:
When using IOMode.once mode, the exception thrown should be a timeout（ReadTimeoutException）, not the default error exception（“Reached end of stream while reading data”）.
I changed the echo-server example to add the "no-pipe" configuration, and I can test it. The original use of once mode, will throw Exception directly, can not determine a read timeout.
 当使用IOMode.once模式时，抛出的异常应该为超时异常，不应该是默认的错误异常。
我更改了 echo-server的例子，添加 “no-pipe” 配置，可以测试。原来使用once模式，会直接抛 Exception，无法判断一次读取超时的。

### DIFF
--- a/examples/echo-server/source/app.d
+++ b/examples/echo-server/source/app.d
@@ -1,20 +1,75 @@
 import vibe.core.core : runApplication;
 import vibe.core.log;
-import vibe.core.net : listenTCP;
+import vibe.core.net;
 import vibe.core.stream : pipe;
 
-void main()
+
+version(READ_WRITE)
 {
-	auto listeners = listenTCP(7000, (conn) @safe nothrow {
-			try pipe(conn, conn);
-			catch (Exception e)
-				logError("Error: %s", e.msg);
-	});
+	import std.exception;
+	import std.datetime;
+	import std.stdio;
+	
+	
+	
+	
+	scope const(ubyte[]) readTcpOnce(ref TCPConnection stream,scope ubyte[] dst){
+		ubyte[] ret = null;
+        collectException!ReadTimeoutException({
+			import eventcore.driver : IOMode;
+			auto len = stream.read(dst, IOMode.once);
+			if(len == 0)
+				return null;
+			else
+				return dst[0..len];
+        }(),ret);
+        return ret;
+	}
 
-	// closes the listening sockets
-	scope (exit)
-		foreach (l; listeners)
-			l.stopListening();
+	void handleTcp(ref TCPConnection stream) @trusted nothrow {
+		auto e = collectException({
+			stream.readTimeout = dur!"seconds"(30);
+			ubyte[] buffer = new ubyte[2048];
+			while(stream.connected){
+				auto tmp = stream.readTcpOnce(buffer);
+				if(tmp is null){
+					writefln("[%s] read time out!",
+        			Clock.currTime);
+				} else {
+					stream.write(tmp);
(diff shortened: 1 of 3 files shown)


source/vibe/core/net.d
```diff
@@ -789,9 +789,16 @@ struct TCPConnection {
...
 		m_context.readTimeout.loopWithTimeout!((remaining) {
 			if (m_context.readBuffer.length == 0) {
-				if (mode == IOMode.immediate || mode == IOMode.once && nbytes > 0)
+				if (mode == IOMode.immediate || (mode == IOMode.once && nbytes > 0))
 					return true;
-				enforce(waitForData(remaining), "Reached end of stream while reading data.");
+				auto ret = waitForDataEx(remaining);
+				if(ret  == WaitForDataStatus.timeout) {
+					// should throw ReadTimeoutException
+					return true;
```
s-ludwig, line 797: Shouldn't this be `throw new ReadTimeoutException("Read operation timed out")` instead?

s-ludwig (changes requested):
It seems like it would make sense to convert the test code in the echo-server example to an actual test case in the tests/ directory. With a short timeout, such as 500 ms and a delayed write on the client side, this would be useful to ensure that the exception, as well as the recovery behavior afterwards, is correct.

source/vibe/core/net.d
dushibaiyu: the "ReadTimeoutException" will throw by loopWithTimeout. it should "return false";
I will fix and add test.

examples/echo-server/source/app.d
```diff
@@ -1,20 +1,74 @@
...
+		}());
+		if(e)
+			collectException({writeln(e.msg);}());
+	}
+	void main()
+	{
+		auto listeners = listenTCP(7000,  (conn) @safe nothrow {handleTcp(conn);});
+
+		// closes the listening sockets
+		scope (exit)
```
s-ludwig, line 50:
```suggestion
		scope (exit)
```

s-ludwig:
Can you also remove the `READ_WRITE` part from the echo server example? Apart from that, everything looks good to merge now.

I'm not sure why the actions are not being run, there is no manual confirmation button for me anymore. But hopefully that will fix itself with the next push.

dushibaiyu:
why not the check run?

s-ludwig:
> why not the check run?

Sorry, I had to manually approve again and didn't notice the new commits.

Now there is still an issue with trailing white space. If you can, I would recommend to get an [editorconfig](https://editorconfig.org/) plugin for your editor. But if you want, I can also check out your branch and fix these things locally, so that this doesn't have to go through the "commit->CI" round trip each time.

dushibaiyu:
ok. I add the "EditorConfig for VS Code". and fix it auto.

s-ludwig:
I actually checked out your branch locally to test the changes and noticed that there is an issue on Windows, where `cancelRead()` results in the connection being closed. For this reason the test would still fail. I'll look into a possible fix.

s-ludwig:
I've pushed a fixed test case in addition to your fix to #366, it skips Windows for now, as I haven't been able to find a fix so far. If that passes, we can close this PR.

dushibaiyu:
thanks。

PR #366, opened 2023-10-18
TCPConnection.read(buffer,IOMode.once) should throw ReadTimeoutException

s-ludwig:
This PR is the squashed version of #359 by @dushibaiyu, with an additional fix for the new test case.


dushibaiyu:
thanks。

PR #374, opened 2023-12-18
Add performInWorker and use it for blocking I/O

s-ludwig:


s-ludwig:
Still need to investigate some rare crashes before this gets merged

s-ludwig:
Added a regression fix for #370 and not appears to be reliable again.

l-kramer (approved):

PR #376, opened 2023-12-19
Fix possible fiber self-call in recycleFiber

s-ludwig:
Fiber.shutdown() calls Fiber.call(), which will crash if the caller is the same fiber. This now lets the fiber exit itself directly if there is no space for recycling. Regression introduced by e877c5b6690a61026e1e651bebe32e69dccea338.


s-ludwig:
Merging without additional review for now, as this is an important hotfix.

PR #382, opened 2024-02-06
Adds meson.build files

denizzzka:


denizzzka:
Perhaps this has lost its relevance - https://github.com/mesonbuild/meson/pull/10045

PR #394, opened 2024-03-02
Use the correct temporary folder on iOS

s-ludwig:
On the device, the /tmp folder is not accessible for regular apps on iOS. `NSTemporaryDirectory` returns a usable directory instead.


l-kramer (approved):

Geod24:
Wouldn't https://dlang.org/library/std/file/temp_dir.html be appropriate ?

s-ludwig:
Could be, I'd have to test whether iOS populates `TMPDIR` appropriately. But for other Posix platforms at least, we should probably use it to make this more robust against more unusual system configurations.

PR #396, opened 2024-03-24
Add utility to build GenericPaths from unencoded strings.

vnayar:
This utility function will be used to url-encode paths added to a URLRouter, which avoids the problem of having to resolve encoding problems when resolving routes.

See: https://github.com/vibe-d/vibe-http/issues/34

### DIFF
--- a/source/vibe/core/path.d
+++ b/source/vibe/core/path.d
@@ -497,6 +497,40 @@ struct GenericPath(F) {
 		return ret;
 	}
 
+	/** Constructs a path from a string which may contain characters that need to be encoded.
+	*/
+	static GenericPath fromUnencodedString(string p) {
+		import std.array : appender;
+		auto dst = appender!string;
+		if (!p.length) {
+			return GenericPath();
+		}
+		string ap = Format.getAbsolutePrefix(p);
+		dst.put(ap);
+		p = p[ap.length..$];
+		while (p.length) {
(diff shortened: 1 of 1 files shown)


s-ludwig:
Hm, the behavior of this function doesn't seem quite sound in the general case. The "unencoded string" is not able to represent all possible `InetPath` paths, but can represent more paths than are possible to represent with a `WindowsPath`. Offering a function like this seems very error prone, as the user would typically not have all edge-cases in mind when using it and the input string format is not clearly defined anywhere.

But, leaving the question of whether this should be a dedicated API aside, isn't this doing the same as `cast(InetPath)PosixPath(str)` when looking at the `InetPath` case?

vnayar:
> But, leaving the question of whether this should be a dedicated API aside, isn't this doing the same as `cast(InetPath)PosixPath(str)` when looking at the `InetPath` case?

I'm still ramping up on this part of the code, not yet fully familiar with its patterns and intent.  The approach of doing `cast(InetPath)PosixPath(str)` seems to do exactly what I was looking for. I'll see if I can use that instead in https://github.com/vibe-d/vibe-http/pull/35 and thereby eliminate the need for this MR.

vnayar:
`cast(InetPath)PosixPath(str)` does exactly what's needed, this MR can be closed.

PR #397, opened 2024-04-01
Ensure uncaught throwables in task fibers are visible to the user

s-ludwig:
Prefers logFatal over stderr.writeln so that the error also ends up in log files that may be configured. Also displays a message box for subsystem "windows" executables that don't have a console attached.


s-ludwig:
The `MessageBox` approach probably needs some refinement prior to merge, as this would not interact well when a subsystem:windows application is run as a non-interactive service.

l-kramer (approved):

PR #399, opened 2024-04-16
Fix lazy evaluation of log arguments

s-ludwig:
Fixes https://github.com/vibe-d/vibe-http/pull/36

### DIFF
--- a/source/vibe/core/log.d
+++ b/source/vibe/core/log.d
@@ -941,8 +941,12 @@ private void doLog(S, T...)(LogLevel level, string mod, string func, string file
     nothrow
 {
 	try {
-		static if(T.length != 0)
-			auto args_copy = args;
+		static if(T.length != 0) {
+			if (!getLoggers().any!(l => l.minLevel <= level))
+				return;
+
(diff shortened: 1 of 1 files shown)


source/vibe/core/log.d
```diff
@@ -941,8 +941,12 @@ private void doLog(S, T...)(LogLevel level, string mod, string func, string file
     nothrow
 {
 	try {
-		static if(T.length != 0)
-			auto args_copy = args;
+		static if(T.length != 0) {
+			if (!getLoggers().any!(l => l.minLevel <= level))
+				return;
+
+			T args_copy = args;
```
Boris-Barboris, line 948: won't compile because if T.length == 0 args_copy is undefined

s-ludwig: But this is inside of `static if (T.length != 0) { ... }`, so it should still work.

Boris-Barboris: Ah, I see, my fork is stuck in old version where formattedWrite is not in static if block

l-kramer (approved):

PR #411, opened 2024-09-10
Fix CI errors and unittest compile error on certain compiler versions

s-ludwig:
Affects the CI of the vibe.d repository

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -21,6 +21,7 @@ jobs:
                 dc: [dmd-latest, ldc-latest]
                 arch: [x86_64]
                 include:
+                  - { os: ubuntu-latest, dc: dmd-2.096.1, arch: x86_64 }
                   - { os: windows-latest, dc: dmd-2.092.0, arch: x86_64 }
                   - { os: windows-latest, dc: dmd-2.092.0, arch: x86_mscoff }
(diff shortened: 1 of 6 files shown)


Geod24:
What's the `in` compiler bug ?

Geod24:
If you can provide me a command line example of compilation and the error message, I can advise.
The obvious reason why you would get a linker error with `in` is if you are compiling a library with a different set of flag than what uses it (with one of those compiling with `-preview=in` but not the other). In which case it would be a feature, not a bug, as the alternative is memory corruption ;)

s-ludwig:
I think it was with dub-registry where I had consistent unresolved symbol errors for all methods that used simple `in` parameters (both, DMD 2.109.1 and LDC 1.39.0). All is compiled in a single `dub build` run, so there should be no compiler argument differences and `-preview=in` shouldn't be involved, unless someone else sneaked it in somewhere. I didn't look for, or write, a bug report so far, because I'm still in a kind of clean-up process after upgrading MongoDB, which caused a little cascade of issues.

Geod24:
https://github.com/dlang/dub/blob/5e2a8de28aa29805dc7db1f6990171f1e1267dab/dub.sdl#L10-L11

s-ludwig:
Okay, I see. I had a quick look, but missed the obvious possibility of `dub`. Would moving those into the "application" configuration do any harm?

l-kramer (approved):

PR #416, opened 2024-11-16
Improves IPv6 addresses formatting

denizzzka:
We have lack of support IPv6 addresses formatting, and especially, IPv4 addresses encapsulated into IPv6

[This code](https://github.com/vibe-d/vibe-http/blob/338ecb4b6aeaf6b9f49a9cd1f550c9d670a1299d/source/vibe/http/server.d#L959) is never called and IPv4 addresses on IPv6 interfaces logged as `0:0:0:0:0:ffff:bca2:ee5` instead of `::ffff:188.162.14.229`

I made own implementation because `inet_ntop` from glibc internally does redundand buffer copying, uses stringz, errno and not handles ~~rare `::2` address case~~

### DIFF
--- a/source/vibe/core/net.d
+++ b/source/vibe/core/net.d
@@ -437,10 +437,6 @@ struct NetworkAddress {
 	/// ditto
 	void toAddressString(scope void delegate(const(char)[]) @safe sink)
 	const nothrow {
-		import std.array : appender;
-		import std.format : formattedWrite;
-		ubyte[2] _dummy = void; // Workaround for DMD regression in master
-
 		scope (failure) assert(false);
 
 		switch (this.family) {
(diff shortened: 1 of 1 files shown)


denizzzka:
`::0.0.0.x` case actually doesn't handled in *nix/Linux, Windows and Postgres due to old bind9 error. In fact, this become a de-facto standard and if this fix, then some comparisons in user code can be unexpectedly breaked

Thus I also removed support for `::0.0.0.x` addresses

s-ludwig:
Sounds like the HTTP server code should really operate on the binary representation of the IPv6 address and then just format the last four bytes as IPv4 for robustness.

But having the special formatting built-in here definitely makes things nicer in other places, too. Can you maybe just also add a simple unit test to `resolveHost` with `use_dns` set to `false`? Just to ensure this is working in both directions for all platforms.

denizzzka:
> Sounds like the HTTP server code should really operate on the binary representation of the IPv6 address and then just format the last four bytes as IPv4 for robustness.

> But having the special formatting built-in here definitely makes things nicer in other places, too. 

Same IPv6 address can be represented by various ways. So we can't rely on any specific format. And there seems to be no such places in Vibe.d

s-ludwig:
> Test that is currently implemented already uses use_dns = false. Added resolveHost call with use_dns = true

Sorry, I've missed that the `toString` tests actually also test the inverse direction.

> Same IPv6 address can be represented by various ways. So we can't rely on any specific format. And there seems to be no such places in Vibe.d

The binary representation is the same (`NetworkAddress.sockAddrInet6.sin6_addr`), but this is something for a separate PR - the changes for `toString` are looking good!

PR #422, opened 2025-01-02
GenericPath cleanup

s-ludwig:
- Swaps the implementation and documentation of `.normalize` and `.normalized`
- Avoids allocations when normalizing an already normalized path
- Deprecates `.Segment2` and associated methods

### DIFF
--- a/source/vibe/core/path.d
+++ b/source/vibe/core/path.d
@@ -723,26 +723,9 @@ struct GenericPath(F) {
 
 	/** Returns the normalized form of the path.
 
-		See `normalize` for a full description.
-	*/
-	@property GenericPath normalized()
-	const {
-		GenericPath ret = this;
-		ret.normalize();
-		return ret;
-	}
-
-	unittest {
-		assert(PosixPath("foo/../bar").normalized == PosixPath("bar"));
-		assert(PosixPath("foo//./bar/../baz").normalized == PosixPath("foo/baz"));
-	}
-
-
-	/** Removes any redundant path segments and replaces all separators by the
-		default one.
-
-		The resulting path representation is suitable for basic semantic
-		comparison to other normalized paths.
+		This removes any redundant path segments and replaces all separators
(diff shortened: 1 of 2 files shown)


source/vibe/core/path.d
```diff
@@ -779,7 +772,30 @@ struct GenericPath(F) {
...
+		assert(PosixPath("/foo/bar").normalized == PosixPath("/foo/bar"));
+		assert(PosixPath("foo/bar/").normalized == PosixPath("foo/bar/"));
+	}
+
+
+	/** Replaces the path representation with its normalized form.
+
+		See `normalized` for a full description.
+	*/
+	void normalize()
```
Geod24, line 796: No deprecation ?

source/vibe/core/path.d
```diff
@@ -779,7 +772,30 @@ struct GenericPath(F) {
...
+		assert(PosixPath("/foo/../bar").normalized == PosixPath("/bar"));
+		assert(WindowsPath(`\\PC/c$\foo/../bar/`).normalized == WindowsPath(`\\PC\c$\bar\`));
+		assert(PosixPath("/foo/bar").normalized == PosixPath("/foo/bar"));
+		assert(PosixPath("foo/bar/").normalized == PosixPath("foo/bar/"));
+	}
+
+
+	/** Replaces the path representation with its normalized form.
+
+		See `normalized` for a full description.
```
Geod24, line 794:
```suggestion
		This is a simple wrapper around `normalized` which should be preferred. See `normalized` for a full description.
```

If you don't want to deprecate it.

Geod24 (approved):
Neat

source/vibe/core/path.d
s-ludwig: I've searched through the code base of a few projects and `normalize()` is being used in a surprising amount of cases, so I'd like to avoid breaking the API for this. I've adjusted the a bit wording here, but went with this solution.

PR #427, opened 2025-05-03
Memory optimization

s-ludwig:


l-kramer (approved):

source/vibe/core/file.d
```diff
@@ -79,11 +79,15 @@ FileStream openFile(string path, FileMode mode = FileMode.read)
...
+	import core.memory : GC;
+
 	auto fil = openFile(path);
 	scope (exit) fil.close();
 	enforce(fil.size <= max_size, "File is too big.");
 	auto sz = cast(size_t)fil.size;
-	auto ret = sz <= buffer.length ? buffer[0 .. sz] : new ubyte[sz];
+	auto ret = sz <= buffer.length
+		? buffer[0 .. sz]
+		: () @trusted { return cast(ubyte[])GC.malloc(sz)[0 .. sz]; } ();
```
Geod24, line 90: Doesn't that mean the GC might scan those ?

s-ludwig: This also means that code using `std.experimental.allocator.gc_allocator` (directly or indirectly) generally has this issue. Seems like a pretty important oversight for the API (just like the inverse issue with `Mallocator`+`make` not calling `GC:addRange` by default).

PR #430, opened 2025-05-13
Improve logException robustness

s-ludwig:
Avoids Exception.toString() and instead logs the diagnostic pieces one-by-one, avoiding dynamic memory allocations and and printing out as much as possible in case of a secondary failure during logging.

Also now never logs the diagnostic part of the exception with a log level higher than the main message.

### DIFF
--- a/source/vibe/core/log.d
+++ b/source/vibe/core/log.d
@@ -227,8 +227,10 @@ void logException(LogLevel level = LogLevel.error)(Throwable exception,
 	string file = __FILE__, int line = __LINE__)
 @safe nothrow {
 	doLog(level, mod, func, file, line, "%s: %s", error_description, exception.msg);
-	try doLog(LogLevel.diagnostic, mod, func, file, line,
-			  "Full exception: %s", () @trusted { return exception.toString(); } ());
+	try () @trusted {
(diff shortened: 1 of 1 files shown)


Geod24:
Well if the `Exception` is given to the formatting function, it should recognize the sink overload: https://github.com/dlang/dmd/blob/0584a56d87a7f8fef57801e1a115cfcae10712f3/druntime/src/object.d#L2667-L2673

You'll have to wrap it in `@trusted` but IMO that's a more correct way to do this.

s-ludwig:
The `@trusted` is necessary for all approaches, so that wouldn't really make a difference. However, another reason for the custom logging was also the redundant output of the exception message and the relatively bad readability of the first line.

I don't know, the main question to me is whether anyone overrides `toString` for this to matter? The fact that the default `toString` performs this elaborate formatting of the various parts doesn't make the overridability appear to be a very inviting or useful facility.

s-ludwig:
With something akin to that `.formatMessage` method instead of the raw `.msg` field, this would be a no-brainer. But you overriding `toString` in practice, with it also extending to dub, is enough evidence - it'll revert that part back to full `toString`.

PR #448, opened 2026-06-07
Add a general DNS query module (vibe.core.dns)

gedaiu:
This adds vibe.core.dns, a general-purpose facility for querying arbitrary DNS record types over UDP. It exposes a DNSRecordType enum covering the full IANA record-type registry and a small, typed API: resolveDNS for raw queries of any type, plus lookupSRV and lookupTXT convenience wrappers. The configured system nameservers are discovered per platform: from /etc/resolv.conf on POSIX (with a fallback to the systemd-resolved stub) and from the network configuration via GetNetworkParams on Windows. The resolver uses a random transaction ID and rejects mismatched replies.

Underneath sits a self-contained, pure wire codec for encoding queries and decoding responses, including DNS name compression. The codec is hardened against malformed and malicious input: out-of-range and cyclic compression pointers, truncated labels, oversized rdlengths, and undersized buffers all raise a catchable exception rather than crashing or looping, and records of unknown types are preserved with their raw rdata instead of being dropped.

The motivation was supporting SRV and TXT lookups for mongodb+srv-style seedlist discovery, but the module is deliberately generic and useful on its own. The wire codec is covered by in-module unit tests, and a network integration test exercises the resolver end-to-end against real SRV and TXT records, skipping cleanly when no network is available.

### DIFF
--- a/source/vibe/core/dns.d
+++ b/source/vibe/core/dns.d
@@ -17,6 +17,8 @@ import std.file : readText, exists;
 import std.random : uniform;
 import std.string : split;
 
+version (Windows) pragma(lib, "iphlpapi");
+
 @safe:
 
 /// Fixed length in bytes of the DNS message header preceding the question section.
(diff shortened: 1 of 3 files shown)


gedaiu:
@s-ludwig can I also get a review on this new dns client?

